PR #23, opened 2015-01-08
Fix a bug in the lookup regexp

xbucchiotty:
Cause a trouble if zookeeper.connection.timeout.ms is present and you want to insert zookeeper.connect


wurstmeister:
Thank you!

PR #33, opened 2015-03-08
Use docker compose instead of fig

kunickiaj:
Updated README to refer to docker-compose instead of fig as it's the replacement.
Also made versions more easily selectable + use closest apache mirror.


probepark:
:+1:

wurstmeister:
Thank you!

PR #41, opened 2015-06-18
fix injection of jvm heap options

varju:
The 0.8.2 release of Kafka moved the KAFKA_HEAP_OPTS export into an if clause, preventing this regex from injecting the external override.


wurstmeister:
Thank you

PR #51, opened 2015-09-03
Enable autoconfigure KAFKA_ADVERTISED_HOST_NAME

jdavisonc:
Enable Kafka to autoconfigure its advertised hostname with local ip address if it is not set.


wurstmeister:
Thank you

PR #57, opened 2015-10-23
Limit waiting time to avoid hanging container when kafka fails to start

eliasdorneles:
When Kafka fails to start (for example, if Zookeeper isn't accessible for some reason), the [loop for waiting it to start](https://github.com/wurstmeister/kafka-docker/blob/4063777543eaef8d18691f2d0fe06915896b1788/start-kafka.sh#L42) can go on forever.

This PR proposes a limit of 10 minutes for that loop.

Does this look good?

### DIFF
--- a/start-kafka.sh
+++ b/start-kafka.sh
@@ -39,7 +39,15 @@ done
 $KAFKA_HOME/bin/kafka-server-start.sh $KAFKA_HOME/config/server.properties &
 KAFKA_SERVER_PID=$!
 
-while netstat -lnt | awk '$4 ~ /:9092$/ {exit 1}'; do sleep 1; done
+wait_timeout=600
+count=0
+while netstat -lnt | awk '$4 ~ /:9092$/ {exit 1}'; do
+    sleep 1;
+    count=$(expr $count + 1)
+    if [ $count -gt $wait_timeout ]; then
+        break
+    fi
+done
 
 if [[ -n $KAFKA_CREATE_TOPICS ]]; then
     IFS=','; for topicToCreate in $KAFKA_CREATE_TOPICS; do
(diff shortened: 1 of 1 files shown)


wurstmeister:
Hi, 

thanks for your PR. I think it would be good to make the timeout configurable and exit with an error message when it is exceeded. 

Thanks

eliasdorneles:
Hi @wurstmeister, thanks for reviewing, sorry for the delay.
It sounds good, I added a commit applying your comments -- does that look good?
Thanks

start-kafka.sh
```diff
@@ -39,7 +39,25 @@ done
 $KAFKA_HOME/bin/kafka-server-start.sh $KAFKA_HOME/config/server.properties &
 KAFKA_SERVER_PID=$!
 
-while netstat -lnt | awk '$4 ~ /:9092$/ {exit 1}'; do sleep 1; done
+
+if [[ -z "$KAFKA_START_WAIT_TIMEOUT" ]]; then
```
wurstmeister: Variables with the KAFKA_ prefix should be reserved for configuration that goes into server.properties. Once we've merged this PR I will update "KAFKA_CREATE_TOPICS" and the documentation

start-kafka.sh
```diff
@@ -39,7 +39,25 @@ done
...
+while netstat -lnt | awk '$4 ~ /:9092$/ {exit 1}'; do
+    sleep 1;
+    count=$(expr $count + 1)
+    if [ $count -gt $KAFKA_START_WAIT_TIMEOUT ]; then
+        start_timeout_exceeded=true
+        break
+    fi
+done
+if $start_timeout_exceeded; then
+    echo "Could not start Kafka broker $KAFKA_BROKER_ID (waited for $KAFKA_START_WAIT_TIMEOUT sec)"
```
wurstmeister: KAFKA_BROKER_ID might be -1 for different brokers (since version 0.9.0.0)

start-kafka.sh
eliasdorneles: Makes sense, updated, thanks!

wurstmeister:
Thank you!

PR #58, opened 2015-11-06
Graceful shutdown of Kafka container.

ethx:
Without this, running `docker-compose stop` doesn't stop Kafka properly (timeout after 10sec and killed -9) and starting again with `docker-compose up -d` results in Kafka issuing errors because previous run didn't unregister from ZK:
```
FATAL Fatal error during KafkaServerStartable startup. Prepare to shutdown (kafka.server.KafkaServerStartable)
java.lang.RuntimeException: A broker is already registered on the path /brokers/ids/1. This probably indicates that you either have configured a brokerid that is already in use, or else you have shutdown this broker and restarted it faster than the zookeeper timeout so it appears to be re-registering.
```


wurstmeister:
Thank you!

PR #61, opened 2015-11-17
Revert "Enable autoconfigure KAFKA_ADVERTISED_HOST_NAME"

kirisetsz:
Reverts commit 4063777543eaef8d18691f2d0fe06915896b1788.

`KAFKA_ADVERTISED_HOST_NAME` shouldn't be autoconfigured since the use case of it is user dependent. For example:

1. Only Kafka instance in docker network
2. Both Kafka instance and client exists in docker network with docker link

Automatically set `KAFKA_ADVERTISED_HOST_NAME` to host address in docker network making it difficult to support the 2nd use case, which is a common case using docker-compose.

### DIFF
--- a/start-kafka.sh
+++ b/start-kafka.sh
@@ -18,10 +18,6 @@ if [[ -n "$KAFKA_HEAP_OPTS" ]]; then
     unset KAFKA_HEAP_OPTS
 fi
 
-if [[ -z "$KAFKA_ADVERTISED_HOST_NAME" ]]; then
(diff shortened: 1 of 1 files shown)


wurstmeister:
Hi, 

thanks for your PR and sorry for the late reply. I've made the hostname command configurable now, so that different environments can easily be supported. (see https://github.com/wurstmeister/kafka-docker/commit/96f98f068f2f626990bc429dfc6e0a17dbb18972)

Thanks

PR #63, opened 2015-11-20
Enable usage in cloud

mikewchicago13:
While running a Kafka cluster in an autoscaling cloud environment, it is not always possible to specify some critical Kafka settings in advance.

- broker.id: This must be unique per instance. When creating a cluster (specifically AWS), it is difficult to specify a unique identifier per server in advance. Therefore, calculate one dynamically if all else fails.
- advertised.host.name: When publishing to Kafka, the publisher must be able to contact a server directly, even if it is behind a load balancer. Therefore, the advertised host name must be addressable by the publisher. In AWS, this can be calculated with a script like ```curl -s http://169.254.169.254/latest/meta-data/public-hostname```

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+* text eol=lf
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,5 @@
 FROM ubuntu:trusty
-
-MAINTAINER Wurstmeister 
+MAINTAINER mweliczko 
 
 ENV KAFKA_VERSION="0.8.2.1" SCALA_VERSION="2.10"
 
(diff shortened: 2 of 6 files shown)


wurstmeister:
Hi, 

thanks for you PR and sorry for the late reply. Based on your suggestion I've made the hostname command configurable now (see https://github.com/wurstmeister/kafka-docker/commit/96f98f068f2f626990bc429dfc6e0a17dbb18972)

When upgrading to 0.9.0.0, the broker id generation should not be necessary anymore (http://kafka.apache.org/documentation.html#upgrade_9_breaking, https://issues.apache.org/jira/browse/KAFKA-1070)

Would you mind updating the PR to only include the supervisor change and a reference to your AWS curl command in the README? 

Thanks

PR #78, opened 2016-03-01
Replace Boot2Docker with Docker Toolbox.

scribu:
Also, add info about setting the advertised host name, since it took me a bit of digging to figure it out.

I just updated the generated `index.html` to highlight the changes I wanted to make. Note that [params.json](https://github.com/wurstmeister/kafka-docker/blob/gh-pages/params.json) is not updated.

I tried to use the "Page Editor" UI from the Github settings, but that [made a mess](https://github.com/scribu/kafka-docker/commit/4147d1904a9db174c39f7b663eb68430282b696b). So yeah, it's not very PR friendly.


cricket007:
Your change is a link for Docker Engine for Mac, not [Docker Toolbox](https://www.docker.com/products/docker-toolbox).

scribu:
You're right. Fixed.

wurstmeister:
Thanks for your PR. I've made the change directly with github pages.

PR #79, opened 2016-03-06
update docker-compose files to use HOSTNAME_COMMAND

zlim:
plus a simple test script


wurstmeister:
Hi, 

thanks for you PR but I don't think there is any reason to use this specific hostname command in the default compose file. we can add it to the readme instead.

PR #84, opened 2016-03-22
Don't store `KAFKA_JMX_OPTS` into server.properties

janantala:
When `KAFKA_JMX_OPTS` env variable is specified it should not be stored into server.properties config file.

### DIFF
--- a/start-kafka.sh
+++ b/start-kafka.sh
@@ -29,12 +29,14 @@ fi
 for VAR in `env`
 do
   if [[ $VAR =~ ^KAFKA_ && ! $VAR =~ ^KAFKA_HOME ]]; then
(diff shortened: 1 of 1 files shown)


choang:
I noticed there are a number of these environment variables used in the `kafka-server-start.sh` script.  Having these values in `server.properties` seems benign as it doesn't cause problems to the server.  I don't think this pr is necessary.

janantala:
@choang `KAFKA_JMX_OPTS` is not part of `server.properties` file, but it defines kafka launch mode
https://github.com/apache/kafka/blob/1e4dd66b19857f4f0ce3f83fd0a808885b0a88c1/bin/kafka-run-class.sh#L137

sscaling:
As per the comment from  @choang. This setting is benign. I don't think the added complexity (i.e. another conditional to debug) is worth what we get in return. Thanks

Issue #17, opened 2014-11-10
Internal host wrong


b0c1:
Hi!

I try to pull up kafka. I created two kafka instance and an another one named kafkaclient (linked to kafka instances). I try to run the 
```
 $KAFKA_HOME/bin/kafka-console-consumer.sh --topic=topic --zookeeper=$ZK_PORT_2181_TCP_ADDR
```

But I got:
```
[2014-11-10 10:11:08,628] WARN [console-consumer-92965_537adcb9a894-1415614259216-39d19909-leader-finder-thread], Failed to find leader for Set([topic,3], [topic,2], [topic,0], [topic,1]) (kafka.consumer.ConsumerFetcherManager$LeaderFinderThread)
kafka.common.KafkaException: fetching topic metadata for topics [Set(topic)] from broker [ArrayBuffer(id:49176,host:8042ab38dd44,port:49176, id:49175,host:df9514a03dca,port:49175)] failed
        at kafka.client.ClientUtils$.fetchTopicMetadata(ClientUtils.scala:67)
        at kafka.client.ClientUtils$.fetchTopicMetadata(ClientUtils.scala:88)
        at kafka.consumer.ConsumerFetcherManager$LeaderFinderThread.doWork(ConsumerFetcherManager.scala:66)
        at kafka.utils.ShutdownableThread.run(ShutdownableThread.scala:51)
Caused by: java.nio.channels.UnresolvedAddressException
        at sun.nio.ch.Net.checkAddress(Net.java:89)
        at sun.nio.ch.SocketChannelImpl.connect(SocketChannelImpl.java:514)
        at kafka.network.BlockingChannel.connect(BlockingChannel.scala:57)
        at kafka.producer.SyncProducer.connect(SyncProducer.scala:141)
        at kafka.producer.SyncProducer.getOrMakeConnection(SyncProducer.scala:156)
        at kafka.producer.SyncProducer.kafka$producer$SyncProducer$$doSend(SyncProducer.scala:68)
        at kafka.producer.SyncProducer.send(SyncProducer.scala:112)
        at kafka.client.ClientUtils$.fetchTopicMetadata(ClientUtils.scala:53)
        ... 3 more

```
It's can't resolve the host address provided by zk. In my custom application if I provide the broker metadata ip/port to the kafka client I got connection refused... any idea why?

eudaimos:
[...]

PR #89, opened 2016-04-20
specify port mapping to avoid random host port

MuhanZou:
Without mapping, we have to run `docker-compose ps` to find out the port of zookeeper, which would be used for $ZK later on. We can avoid this little annoying thing by explicitly specifying the port mapping. #17


wurstmeister:
please see https://github.com/wurstmeister/kafka-docker/blob/master/docker-compose-single-broker.yml

PR #104, opened 2016-06-21
Replication factor is 3 by default, preventing single-broker

rberenguel:
Kafka defaults to replication factor of 3 for the topic __consumer_offsets (even if __consumer_offsets:1:1 is created on launch) preventing single-broker from accepting any consumers (since the number of replicas won't match)

### DIFF
--- a/docker-compose-single-broker.yml
+++ b/docker-compose-single-broker.yml
@@ -12,5 +12,6 @@ services:
       KAFKA_ADVERTISED_HOST_NAME: 192.168.99.100
       KAFKA_CREATE_TOPICS: "test:1:1"
       KAFKA_ZOOKEEPER_CONNECT: zookeeper:2181
+      KAFKA_OFFSETS_TOPIC_REPLICATION_FACTOR: 1
     volumes:
       - /var/run/docker.sock:/var/run/docker.sock


wurstmeister:
Hi, 

can you please describe the problem you are having a bit more, i.e. steps to reproduce, versions used, etc? 

running ```docker-compose -f docker-compose-single-broker.yml up``` against the current master allows me to connect the console consumer and producer and exchange messages.

AFAICT https://issues.apache.org/jira/browse/KAFKA-1846 addresses the problem you are describing but without exact steps to reproduce it's hard to tell (also see https://github.com/apache/kafka/blob/79aaf19f24bb48f90404a3e3896d115107991f4c/core/src/main/scala/kafka/server/KafkaApis.scala#L648)

rberenguel:
I also saw the fixed issue, and the documentation about min(alive, 3), but somehow since it was still failing I tried to force it, and it fixed it. But indeed, I'm increasingly puzzled:

I just tried a sample docker-compose as the one I was using and a consumer connects without issues. For a few minutes I have toyed with the following timeline for the issue on my side:
* I started kafka with scale=3, hence topic replication was set to 3
* Connected a consumer (python-kafka)
* Shut down kafka (trying to fix issues with a producer and the networking in compose)
* The consumer kept the topic replication as expecting 3 when connecting, making a single broker fail (this is not how I understand Kafka works, though, but I ran out of reasonable ideas anyway)
I'm definitely puzzled right now, because this should not have happened (when doing all the tries I restarted the consumer and everything should have picked the correct replication factor instead). Also, when I did the "fix" it started working properly, which also should not have happened because it was never supposed to have a 3 (unless the scale=3 was affecting at some other level instead).

Definitely weird. I'm closing the request, it seems to definitely not be needed. But well, if someone in the future has some weird issue with the replication factor they will have some more Google result for it!

Thanks for taking the time to checking it and linking to the specific line in the sources, I didn't dig that deep (and also thanks for creating and sharing this machine).

PR #113, opened 2016-08-08
Simplify download process + some cleaning work

iemejia:
I removed the extra shell script to download kafka and removed some unused deps.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,16 +1,17 @@
 FROM anapsix/alpine-java
 
-MAINTAINER Wurstmeister 
+MAINTAINER Wurstmeister
 
-RUN apk add --update unzip wget curl docker jq coreutils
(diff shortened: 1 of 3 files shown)


wurstmeister:
Hi,
 
the current download process picks the closest mirror. why would this need to be removed?

Thanks

iemejia:
You are right, I should probably return back to the closest mirror version, I just wanted to make the code simpler and get rid of the .sh file, if you agree I can rebase it.

sscaling:
Closing due to being significantly behind master.

The only thing I can see that this currently adds is the AWS hostname command

```
-HOSTNAME_COMMAND=wget -t3 -T2 -qO-  http://169.254.169.254/latest/meta-data/local-ipv4
+HOSTNAME_COMMAND=curl http://169.254.169.254/latest/meta-data/local-ipv4
```

Please feel free to reopen / rebase for updating that command. Thanks.

PR #117, opened 2016-08-23
fix for alpine linux repositories

holinov:
because of some troubles with package repositories original image stopped to work
this fix enables direct access to repository if CDN could-not resolve target host


wurstmeister:
Hi, i don't see any issues with the package repo. If it was just a temporary issue with the CDN I don't think we need this fix.

PR #121, opened 2016-09-12
Bash script wait for KAFKA process

yanndegat:
The container has to terminate if the java process terminates; 
Not waiting for the java process is making the container wait 
for the "create-topic.sh" timeout. even if there's no kafka process.


wurstmeister:
Thank you!

PR #125, opened 2016-09-20
The scripts need to have executable permission

Writtic:
Hi! I'm happy to use the repository for my project. but I found some slight error.
The scripts need to have executable permission especially on Window.
I suppose docker on any other OS is find.
thx 😄

And FYI about this request, take a look this docker/compose#2301


wurstmeister:
Thanks you!

PR #143, opened 2016-12-02
Don't default ADVERTISED_PORT if listeners are configured

pbrisbin:
If configuring listeners (and/or advertised.listeners) to support a
configuration with both PLAINTEXT and SSL endpoints[1], you must not set either
of advertised.{host,port}, otherwise the broker will only create the PLAINTEXT
endpoint.

[1]: http://docs.confluent.io/2.0.0/kafka/incremental-security-upgrade.html

I wasn't sure how to format this, so please let me know if you prefer it
differently. Also, would it be possible to back-port this to the latest 0.9
tag?

### DIFF
--- a/start-kafka.sh
+++ b/start-kafka.sh
@@ -3,7 +3,9 @@
 if [[ -z "$KAFKA_PORT" ]]; then
     export KAFKA_PORT=9092
 fi
-if [[ -z "$KAFKA_ADVERTISED_PORT" ]]; then
+if [[ -z "$KAFKA_ADVERTISED_PORT" && \
+  -z "$KAFKA_LISTENERS" && \
+  -z "$KAFKA_ADVERTISED_LISTENERS" ]]; then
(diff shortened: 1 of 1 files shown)


start-kafka.sh
```diff
@@ -3,7 +3,9 @@
 if [[ -z "$KAFKA_PORT" ]]; then
     export KAFKA_PORT=9092
 fi
-if [[ -z "$KAFKA_ADVERTISED_PORT" ]]; then
+if [[ -z "$KAFKA_ADVERTISED_PORT" && \
+  -n "$KAFKA_LISTENERS" && \
+  -n "$KAFKA_ADVERTISED_LISTENERS" ]]; then
```
pbrisbin: Just realized, I think I intended these to be `-z`: don't default the `_PORT` unless none of these values are present. Will update.

pbrisbin:
Ping @wurstmeister do you think you'll be able to review this? Sorry to bother, just want to know if we should start thinking about keeping our own images for this change.

wurstmeister:
Thank you!

PR #145, opened 2016-12-13
Add log compaction support

d33d33:
Edit create-topic to support a fourth argument handling cleanup.policy config
Update Readme


wurstmeister:
Thank you

PR #157, opened 2017-01-25
Using args in Dockerfile to make versioning easier

jzakrzeski:
- default to latest stable kafka (0.10.1.1) and scala (2.11)


wurstmeister:
Thank you!

PR #158, opened 2017-01-26
Bump Kafka to 0.10.1.1 stable

nicholasserra:
Bump to the 0.10.1.1 release of Kafka.

Built the container with these changes locally and looks good.


nicholasserra:
Looks like this same change landed in #157 thanks!

PR #160, opened 2017-01-28
Change how create-topics.sh checks for running Kafka

dead10ck:
It seems in Kafka 0.8, the port is bound early in the process of starting up the server (I saw this happening in `network_mode: host`, which may have something to do with it), so sometimes `create-topics.sh` will proceed to start creating the topics before the Kafka server is up, which causes errors. If the broker ID is set, then a perhaps more reliable way to check that the server is up and running is to check that the ephemeral ZooKeeper node `/brokers/ids/<ID>` exists. This is the node that Kafka uses to check the health of other brokers in the cluster, so if the node is there, then that indicates the broker is up and healthy. If the broker ID is not set, then unfortunately, there doesn't appear to be a reliable way to check which broker ID was assigned, so checking for port binding is a reasonable fallback. Maybe another option in this case is just grepping the server logs for a line that looks like

```
[2017-01-28 20:57:19,354] INFO [Kafka Server 0], started (kafka.server.KafkaServer)
```

which is kind of hacky, but may still be more reliable.

### DIFF
--- a/create-topics.sh
+++ b/create-topics.sh
@@ -8,7 +8,17 @@ fi
 start_timeout_exceeded=false
 count=0
 step=10
-while netstat -lnt | awk '$4 ~ /:'$KAFKA_PORT'$/ {exit 1}'; do
+
+check_kafka_running() {
+    if [[ "$KAFKA_BROKER_ID" != "-1" ]]; then
+        echo dump | nc -w 1 "$KAFKA_ZOOKEEPER_CONNECT" $KAFKA_PORT \
+            | grep -F "/brokers/ids/$KAFKA_BROKER_ID"
+    else
+        netstat -lnt | awk '{ print $4 }' | grep ":$KAFKA_PORT"
+    fi
+}
+
+while ! check_kafka_running; do
     echo "waiting for kafka to be ready"
     sleep $step;
     count=$(expr $count + $step)


dead10ck:
Paging @wurstmeister ?

wurstmeister:
Hi @dead10ck,

another option would be to use kafkacat 

```
check_kafka_running() {
     [[ `kafkacat -L -b localhost:$KAFKA_PORT  -J  | jq '.brokers| length'` -gt "0" ]]
}
```

I've not had a chance to test this properly yet though

dead10ck:
@wurstmeister I think it should still check for the broker ID, though, since we want to allow for multiple brokers to be running. If we don't check for the container's specific assigned broker ID, then that check would not work after the first broker is up.

wurstmeister:
Hi @dead10ck 

i've created an example using kafkacat https://github.com/wurstmeister/kafka-docker/blob/CREATE_TOPIC/create-topics.sh

The main idea is that, if we can read the metadata from a broker, the broker must be up. this also allows us to check if we actually have enough brokers available to create a given topic.

you can test this by running 

```docker-compose up``` 

this will create ```Topic1``` as we will have one broker available. if you run ```docker-compose scale kafka=2```  ```Topic2``` will be created as well, as soon as the second broker is up.

**Comment by dead10ck** [comment]
@wurstmeister Oh, I see; ok, I'll try this out and let you know when I've updated the PR.

**Comment by dead10ck** [comment]
@wurstmeister it looks like `kafkacat` can't connect to the broker:

```sh
bash-4.3# /usr/lib/jvm/default-jvm/bin/jps -l
1 kafka.Kafka
569 sun.tools.jps.Jps
bash-4.3# kafkacat -L -b 127.0.0.1 -J
% ERROR: Failed to acquire metadata: Local: Timed out
```

Not sure what's going on with it, but the broker is up, so I'm not sure that the `kafkacat` tool works.

I also noticed in this branch that you added a line in the `Dockerfile` that uninstalls `wget` and `curl`, which the `download-kafka.sh` script needs, so the build doesn't work.

PR #162, opened 2017-01-30
jmx-support: Fix start script so sub-process is not created

sscaling:
This makes environment variables available to the kafka-server-start.sh script. This allows JMX_PORT
and other options to be set, and allow the correct configuration of JMX

Change-Id: If1434b5a3fb9d094cd5943656e8a38d9bc60d1e7


sscaling:
Also, it appears a different approach was taken here: https://github.com/wurstmeister/kafka-docker/commit/a6dbce0e71945e77bc60980422183e5c0d454bd8

But was removed in a subsequent commit: https://github.com/wurstmeister/kafka-docker/commit/9535dc1b8b9cfdf519435e1c6f48351c49817529

wurstmeister:
Thank you!

PR #168, opened 2017-02-11
docker-compose file for Docker Swarm mode

screwgoth:
docker-compose file that will work with Docker stack deployment as per Docker v1.13 and above.
Due to the load-balancing and service discovery features  of Docker Swarm mode in Docker v1.13, we can safely expose the Kafka port and point the ADVERTISED_HOST to any one of the docker hosts in the Swarm cluster.

### DIFF
--- a/docker-compose-stack.yml
+++ b/docker-compose-stack.yml
@@ -1,0 +1,16 @@
+version: '3'
+services:
+  zookeeper:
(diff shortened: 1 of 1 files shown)


sscaling:
Thanks for the PR. However, PR #199 supersedes this - please feel free to review and create a new PR if any further amendments are needed.

PR #176, opened 2017-04-09
Use --partitions in create-topics.sh

iamdanfox:
It seems the create-topics.sh script is failing for me on 0.10.1.1, with the following error:

    Missing required argument "[partitions]"
    kafka_1           | Option                                   Description
    ...

I think this can be fixed by changing --partition to --partitions.  This matches the [kafka docs][1].

[1]: https://kafka.apache.org/documentation/

For reference, here is my docker-compose.yml

```
version: '2'
services:
  zookeeper:
    image: wurstmeister/zookeeper
    ports:
      - "2181:2181"
  kafka:
    image: wurstmeister/kafka:0.10.1.1
    ports:
      - "9092:9092"
    environment:
      KAFKA_ADVERTISED_HOST_NAME: kafka
      KAFKA_ADVERTISED_PORT: 9092
      KAFKA_ZOOKEEPER_CONNECT: zookeeper:2181
      KAFKA_CREATE_TOPICS: "resource_smoke_test:1,2"
```

### DIFF
--- a/create-topics.sh
+++ b/create-topics.sh
@@ -28,9 +28,9 @@ if [[ -n $KAFKA_CREATE_TOPICS ]]; then
         echo "creating topics: $topicToCreate"
         IFS=':' read -a topicConfig <<< "$topicToCreate"
         if [ ${topicConfig[3]} ]; then
-          JMX_PORT='' $KAFKA_HOME/bin/kafka-topics.sh --create --zookeeper $KAFKA_ZOOKEEPER_CONNECT --replication-factor ${topicConfig[2]} --partition ${topicConfig[1]} --topic "${topicConfig[0]}" --config cleanup.policy="${topicConfig[3]}"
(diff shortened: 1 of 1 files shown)


wurstmeister:
Hi, 

sorry for the late reply.
 
both ```--partition``` and ```--partitions``` work, even though you are correct that ```--partitions``` is the documented option.

```
bash-4.3# kafka-topics.sh --create --zookeeper zookeeper  --partition 1  --topic test --replication-factor 1
Created topic "test".
bash-4.3# kafka-topics.sh --create --zookeeper zookeeper  --partitions 1  --topic test2 --replication-factor 1
Created topic "test2".
```

the problem you have is the wrong format for ```KAFKA_CREATE_TOPICS```
it should be: ```KAFKA_CREATE_TOPICS: "resource_smoke_test:1:2"```

Please let me know if this fixes your issue, then I will merge your PR to use the documented option

Thanks

tcjennings:
Looks like Kafka 0.11.0.0 no longer supports "--partition" and only supports "--partitions"

PR #179, opened 2017-04-15
Accept BROKER_ID_COMMAND env var for setting broker.id

dominikkrejcik:
BROKER_ID_COMMAND works the same way as HOSTNAME_COMMAND, allowing you
to set broker.id via a command that gets evaluated by the startup
script.

This is useful for extracting the broker.id from the hostname when running in environments like Kubernetes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,6 +37,11 @@ The default ```docker-compose.yml``` should be seen as a starting point. By defa
 
 ## Broker IDs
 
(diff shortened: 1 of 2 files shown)


wurstmeister:
Hi, 

could you please also update the readme and add an example. 

Thanks

wurstmeister:
Thank you!

PR #180, opened 2017-04-21
DOCKER_HOST_IP Enironment Variable

hmusavi:
Added DOCKER_HOST_IP environment variable to docker_compose file to avoid hard-coding them and not require changes to yml file for user environment. Also added two convenience scripts to the kafka image.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,6 @@
 FROM anapsix/alpine-java
 
-ARG kafka_version=0.10.1.1
+ARG kafka_version=0.10.0.1
 ARG scala_version=2.11
 
 MAINTAINER wurstmeister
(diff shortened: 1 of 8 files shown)


sscaling:
Hi @hmusavi, thanks for the PR. Could you provide a bit more context on the issue this is solving and perhaps the use cases that it addresses. I'm just trying to get my head around this change. 

The `docker-compose.yml` files are examples and, in general, should be configured per environment. Typically you'd have a kubernetes/marathon/swarm configuration files for a specific deployment.

hmusavi:
My goal was to make the whole configuration independent of the user environment so one can use it out of the box without having to modify the docker-compose or anything. All one has to do is define the right environment variable (or a .env) file and use the image straight from DockerHub even. For instance, instead of having the local ip address of the host in the docker-compose, if we have use an environment variable, it achieves that goal. Then I ended up using the same variable in a couple of scripts as well. Now that we have it, why not. 

Additionally, I just added the two kafka console scripts to the docker image as well because I found them quite handy having them around in the image. I use them all the time for troubleshooting and figured other may find them useful too.  

So, as you can see, these are very trivial changes that I have deployed in my fork. If you approve and merge them, then I won't have to keep a separate fork. I can use your image out of the box. But if you feel that these aren't necessary, I understand ; feel free to ignore them. It won't hurt my feeling :-)

PR #188, opened 2017-05-12
map LOG4J_% env vars to entries in log4j.properties

### LINKED ISSUE #120: Configuring log4j without touching the image
The `KAFKA_` environment variables handled by the startup script are very handy for tweaking Kafka without touching the image itself. I'm looking for a similar way of configuring log4j, particularly to set a more verbose log level in some cases.

Would it be worthwhile setting up `LOG4J_` 👉  `log4j.properties` as well, so that it works similar to `KAFKA_` and `server.properties`?

(https://github.com/wurstmeister/kafka-docker/blob/master/start-kafka.sh#L29-L40)

Note: This is my first time dealing with log4j, so I'm very open to suggestions on a better way to configure log4j.

gwilym:
I took the existing KAFKA_% env var map for config.properties and applied it to log4j.properties

Seems to work for me for tweaking log4j outputs/verbosity (particularly debugging the kafka authorizer), but I'm not very familiar with log4j so it may not work for absolutely every situation

e.g.

```
LOG4J_LOGGER_KAFKA_AUTHORIZER_LOGGER=DEBUG, authorizerAppender
```

fixes #120


wurstmeister:
Thanks very much. Could you please add some information to the README as well? 

Thanks

gwilym:
Hi @wurstmeister, I've pushed up a readme change which hopefully explains it clearly.

Cheers.

wurstmeister:
Thank you!

Issue #146, opened 2016-12-20
Build image does not work on Windows 10 - /tmp/download-kafka.sh: not found


m-a-d-e:
Hello,
When I run the build command on Windows 10, Docker 1.12.3, I get the following error:

> Step 6 : RUN chmod a+x /tmp/download-kafka.sh && sync && /tmp/download-kafka.sh && tar xfz /tmp/kafka_${SCALA_VERSION}-${KAFKA_VERSION}.tgz -C /opt && rm /tmp/kafka_${SCALA_VERSION}-${KAFKA_VERSION}.tgz && ln -s /opt/kafka_${SCALA_VERSION}-${KAFKA_VERSION} /opt/kafka
>  ---> Running in f32b8949e634
> /bin/sh: /tmp/download-kafka.sh: not found
> ERROR: Service 'kafka' failed to build: The command '/bin/sh -c chmod a+x /tmp/download-kafka.sh && sync && /tmp/download-kafka.sh && tar xfz /tmp/kafka_${SCALA_VERSION}-${KAFKA_VERSION}.tgz -C /opt && rm /tmp/kafka_${SCALA_VERSION}-${KAFKA_VERSION}.tgz && ln -s /opt/kafka_${SCALA_VERSION}-${KAFKA_VERSION} /opt/kafka' returned a non-zero code: 127

I tried to split the "chmod" and the execution of the script into 2 distinct commands but that did not fixed the problem.
I started from a clean situation (no images, no (ended) containers running.

m-ciesielski:
Hello @m-a-d-e ,
I've got the same error on Windows 10 Docker installation.  I've tried to run `download-kafka.sh` script manually on container and following error message was printed:
```
λ docker run -it 9652fded8cea /bin/bash
bash-4.3# cd /tmp
bash-4.3# ./download-kafka.sh
bash: ./download-kafka.sh: /bin/sh^M: bad interpreter: No such file or directory
bash-4.3# bash download-kafka.sh
download-kafka.sh: line 2: $'\r': command not found
```
So issue was caused by Windows CRLF line endings in `download-kafka.sh` script - after removing them (with `dos2unix` in my case), I was able to build kafka-docker image without further problems.

Cyb3rFl3x:
OK... i had the same Problem on Mac with Zsh and i found a workaround:
The workaround was to run the script with bash.
So use: /bin/bash ./start-kafka-shell.sh

PR #195, opened 2017-06-13
Update the Dockerfile to work on Windows machines

auggiecc87:
This is a relatively simple change to allow people using windows to not get hosed when they pull this repo (I know, "but why are you using windows?", maybe because that is what my company uses).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,11 +5,11 @@ ARG scala_version=2.12
 
 MAINTAINER wurstmeister
 
-RUN apk add --update unzip wget curl docker jq coreutils
+RUN apk add --update unzip wget curl docker jq coreutils dos2unix --update-cache --repository http://dl-3.alpinelinux.org/alpine/edge/testing/ --allow-untrusted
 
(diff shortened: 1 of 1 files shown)


auggiecc87:
I understand that there is a setting on git that allows the checkouts (and checkins) to keep linux line endings, but that would require a check in for testing, and/or only pulling with no editing.

I think this is a great way to work around people using windows to edit / add files, and give them the template to add more scripts. I leave for reference this article which basically says: "Yeah, it kinda sucks but this is the best work around still"

http://willi.am/blog/2016/08/11/docker-for-windows-dealing-with-windows-line-endings/

sscaling:
Fixes https://github.com/wurstmeister/kafka-docker/issues/146

sscaling:
Potentially fixes https://github.com/wurstmeister/kafka-docker/issues/262

odin88:
Not working for me.  Using Windows 10. I clone this repo and change the Dockerfile based on @auggiecc87 fix but it is not working

PR #225, opened 2017-10-27
Compact Dockerfile (many fewer layers)

extemporalgenome:
This Dockerfile produces an image with fewer layers; ENV, COPY, and RUN sections do not need to be separate, and are consolidated.


thall:
Looking good!

wurstmeister:
Thank you!

PR #226, opened 2017-10-27
Create topics in parallel

extemporalgenome:
Serial topic creation may require over a minute to create just a few dozen topics. This PR creates topics in parallel.

On my machine, 32 topics take 1m35s to create serially, but 10-15s to create in parallel (both figures including kafka daemon initialization time).

Also, don't bother waiting for Kafka to finish starting if there are no topics to create.


wurstmeister:
Thank you!

Issue #211, opened 2017-08-21
WARN Connection to node 1001 could not be established. Broker may not be available.


SquirrelNinja:
Running this on Mac Sierra

Started this up using this config:

```
version: '2'
services:
  zookeeper:
    image: wurstmeister/zookeeper
    ports:
      - "2181:2181"
  kafka:
    build: .
    ports:
      - "9092"
    environment:
      KAFKA_ADVERTISED_HOST_NAME: 192.168.99.100
      KAFKA_ZOOKEEPER_CONNECT: zookeeper:2181
      KAFKA_CREATE_TOPICS: "lambda1:1:1"
    volumes:
      - /Users/<user>/projects/aa_dev_kafka/docker.sock:/var/run/docker.sock
```

Everything appears to start up, however the log kafka log is full of these errors:

WARN Connection to node 1001 could not be established. Broker may not be available. (org.apache.kafka.clients.NetworkClient)

I saw a couple of post indicating it might be a firewall issue, however I am not running a FW locally

SquirrelNinja:
I deleted the images, the local mapping, and set the docker-compose.yml file to it's stock setting and I am still getting the error.

agolomoodysaada:
I'm also experiencing this issue with
```yml
version: '3.3'
services:
  kafka:
    image: wurstmeister/kafka:0.11.0.0
    ports:
      - 9092:9092
    network_mode: host
    environment:
      KAFKA_ADVERTISED_HOST_NAME: kafka
      KAFKA_AUTO_CREATE_TOPICS_ENABLE: 'true'
      LOG4J_LOGGER_KAFKA_AUTHORIZER_LOGGER: DEBUG, authorizerAppender
  zookeeper:
    image: zookeeper:3.4
    ports:
      - 2181:2181
    network_mode: host
    environment:
      ZOO_MY_ID: 1
```

Eoan242:
I am new to this docker image so I assumed it had to be me doing it wrong. For the past two days I have been trying to get this running on AWS ECS. No matter what I try with every docker-compose up, it seems that the kafka broker alternates between shutting down or throwing this "warning" indefinitely.

nickdani:
Have the same issue :(

SciTechPilgrim:
I was able to Resolve this issue on my setups. It appears that this behavior is Kafka correctly reporting that it is unable to open a TCP connection to the Docker Host. There must be code changes in Kafka 0.11.0 which make this more likely to occur.

[...]

PR #227, opened 2017-10-31
It should fix issue #211

camelaissani:


wurstmeister:
please see https://github.com/wurstmeister/kafka-docker/blob/master/docker-compose-single-broker.yml#L10

the port is configured like this so we can use ```docker scale```

camelaissani:
Okay it makes sense

Issue #233, opened 2017-11-15
KAFKA_ADVERTISED_HOST_NAME


cmreigrut:
According to the docs, setting KAFKA_ADVERTISED_HOST_NAME in the environment section should change the advertised host, right?  It doesn't appear to work...

```
docker-compose.yml
version: '2'
services:
  zookeeper:
    image: wurstmeister/zookeeper
    ports:
      - "2181:2181"
  kafka:
    image: wurstmeister/kafka
    ports:
      - "9092:9092"
    environment:
      KAFKA_ZOOKEEPER_CONNECT: zookeeper:2181
      KAFKA_ADVERTISED_HOST_NAME: 10.17.230.67
      KAFKA_CREATE_TOPICS: "test:1:1"
    volumes:
      - /var/run/docker.sock:/var/run/docker.sock
```

log from docker-compose up
```
kafka_1      | [2017-11-15 00:14:49,290] INFO KafkaConfig values:
kafka_1      |  advertised.host.name = null
kafka_1      |  advertised.listeners = PLAINTEXT://10.17.230.67:9092
kafka_1      |  advertised.port = null
kafka_1      |  alter.config.policy.class.name = null
kafka_1      |  authorizer.class.name =
kafka_1      |  auto.create.topics.enable = true
```

sscaling:
Correct. We're backing out a change that was a little too prescriptive and allowing the configuration as specified by the environment. After #286 is merged you will see `advertised.host.name` configured as you expect.

Please follow that for progress.

PR #236, opened 2017-11-28
Fix deprecated advertised host name

trajakovic:
Update readme and docker-compose* files.

This addresses #233, and probably fixes latest #211 errors.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@ The image is available directly from [Docker Hub](https://hub.docker.com/r/wurst
 ## Pre-Requisites
 
 - install docker-compose [https://docs.docker.com/compose/install/](https://docs.docker.com/compose/install/)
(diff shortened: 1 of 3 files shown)


trajakovic:
although this fixes startup issue, kafka brokers don't work outside docker containers

trajakovic:
closing this as it's not complete solution

AndrewOrr:
I think it is a problem with the advertised hostname? When you use the boot2docker ip I can connect outside of the docker container.

tcjennings:
The reason this doesn't work is changing "KAFKA_ADVERTISED_HOST_NAME" TO "KAFKA_ADVERTISED_LISTENERS" is the same as just removing "KAFKA_ADVERTISED_HOST_NAME" entirely.

The resulting broker configures a single listener on port 9092 that will work within Docker but not from outside. To create a broker that will from outside Docker, a more robust listener configuration has to be configured. Specifically, you need 2 listeners, one for inside and one for outside, on different ports, and publish only the outside listener to your host.

tcjennings:
See PR #256

PR #256, opened 2018-01-28
Kafka-Docker SSL and Etc

tcjennings:
Main feature in this PR is the addition of support for SSL-secured brokers in the container. README details necessary steps to setup SSL and describes the operation of the new ssl-boostrap.sh script that generates self-signed certificates in the absence of other configuration.

Other features include

* Use Java 9 base image which is supported by Kafka 1.0.0 and improves performance of  TLS operations as noted in the 1.0.0 release notes.
* Simplify the Kafka download process in the Dockerfile; eliminate separate download script and dependence on wget and jq.
* Add Label Schema (http://label-schema.org/rc1/) LABELs to Dockerfile to influence build metadata.
* Include SSL and GSSAPI packages (~12MB to final image size) to support SSL and potentially SASL in container.


tcjennings:
Added update to the default compose file that more properly configures a broker for single-broker single-host operation that works from both inside and outside Docker.

tcjennings:
Addresses #211

tcjennings:
Reverted to the Java 8 base image after observing some topic compression issues in the field that I could not replicate in testing. Will revisit the Java 9 base when time allows.

sscaling:
Hi @tcjennings  - it would be great if this change solely focuses on SSL (I think we should pull out labels, other optimizations (such as removal of `download-kafka.sh`), dependency thinning (removal of wget etc) into separate PRs to keep this change as concise and focused as possible.

I'm happy to raise those (although, I have a `download-kafka.sh` PR open #270 for if the mirror is unavailable, which may be worth keeping) so you can rebase. What do you think?

ssl-bootstrap.sh
```diff
@@ -0,0 +1,109 @@
...
+fi
+
+if [[ -f /run/secrets/server.truststore.jks ]]; then
+   export KAFKA_SSL_TRUSTSTORE_LOCATION=/run/secrets/server.truststore.jks
+fi
+
+#If the Keystore file already exists (i.e., has been volume-mounted or loaded as a secret) don't run this script any more
+if [[ ! -f ${KAFKA_SSL_KEYSTORE_LOCATION} ]]; then
+
+if [[ -z "$CERT_DN" ]]; then
```
sscaling: Should this whole block be indented so that the nesting is apparent?

ssl-bootstrap.sh
```diff
@@ -0,0 +1,109 @@
...
+export CA_KEY=${CA_KEY-/etc/ssl/private/ca_key} #The CA Private key used for signing
+export CA_PASSWORD=${CA_PASSWORD-password1234} #The CA Private key password
+export CA_P12_FILE=${CA_P12_FILE-} #Use an outside CA key pair as a PKCS12 file
+export CA_P12_PASSWORD=${CA_PA12_PASSWORD-password1234} #the password for the P12 file
+
+cd /etc/ssl/private
+
+#Check if the Keystore and/or Truststore files are available as Secrets; if they are, then
+#this script does not need to run.
+if [[ -f /run/secrets/server.keystore.jks ]]; then
```
sscaling: It might be better to make it user-explicit which keystore and truststore they are using, rather than automatically setting it. Especially as I think this is just for swarm secrets

ssl-bootstrap.sh
```diff
@@ -0,0 +1,109 @@
...
+
+#If the Keystore file already exists (i.e., has been volume-mounted or loaded as a secret) don't run this script any more
+if [[ ! -f ${KAFKA_SSL_KEYSTORE_LOCATION} ]]; then
+
+if [[ -z "$CERT_DN" ]]; then
+export CERT_DN="cn=$(hostname), ou=GenericOU, o=GenericO, c=US"
+fi
+
+if [[ -z "$CA_DN" ]]; then
+export CA_DN="/C=US/ST=California/L=Cupertino/CN=$(hostname)"
```
sscaling: Some  ifs are indented, others are not - could we make it consistent?

ssl-bootstrap.sh
```diff
@@ -0,0 +1,109 @@
...
+if [[ -n "${CA_P12_FILE}" ]]; then
+openssl pkcs12 -in ${CA_P12_FILE} -out ca-cert -nokeys -nodes -passin pass:${CA_P12_PASSWORD}
+openssl pkcs12 -in ${CA_P12_FILE} -out ca-key -nocerts -nodes -passin pass:${CA_P12_PASSWORD}
+fi
+
+echo "##########"
+echo "Importing Root CA certificate to Client and Server Truststores."
+
+keytool -keystore $KAFKA_SSL_TRUSTSTORE_LOCATION -alias CARoot -import -file ca-cert -storepass ${KAFKA_SSL_TRUSTSTORE_PASSWORD} -noprompt
+#keytool -keystore client.truststore.jks -alias CARoot -import -file ca-cert -storepass ${KAFKA_SSL_TRUSTSTORE_PASSWORD} -noprompt
```
sscaling: Can this comment be removed?

Dockerfile
```diff
@@ -3,21 +3,28 @@ FROM anapsix/alpine-java
...
- && mv /tmp/start-kafka.sh /tmp/broker-list.sh /tmp/create-topics.sh /usr/bin \
- && /tmp/download-kafka.sh \
- && tar xfz /tmp/kafka_${SCALA_VERSION}-${KAFKA_VERSION}.tgz -C /opt \
- && rm /tmp/kafka_${SCALA_VERSION}-${KAFKA_VERSION}.tgz \
+RUN apk add --update unzip curl docker coreutils cyrus-sasl cyrus-sasl-gssapi krb5 openssl \
+ && chmod a+x /usr/bin/*.sh \
+ && curl -L "https://www.apache.org/dyn/closer.cgi?action=download&filename=/kafka/${KAFKA_VERSION}/kafka_${SCALA_VERSION}-${KAFKA_VERSION}.tgz" \
+  | tar xzf - -C /opt \
  && ln -s /opt/kafka_${SCALA_VERSION}-${KAFKA_VERSION} /opt/kafka \
  && rm /tmp/*
```
sscaling: This fails for me locally

```
rm: cannot remove '/tmp/*': No such file or directory
```

ssl-bootstrap.sh
```diff
@@ -0,0 +1,109 @@
...
+export CA_DN="/C=US/ST=California/L=Cupertino/CN=$(hostname)"
+fi
+
+echo "##########"
+echo "Creating a Keystore with a Private Key for the Broker"
+
+keytool -keystore ${KAFKA_SSL_KEYSTORE_LOCATION} -alias localhost -validity 365 -keyalg RSA -genkey \
+-storepass ${KAFKA_SSL_KEYSTORE_PASSWORD} \
+-dname "$CERT_DN" \
+-keypass ${KAFKA_SSL_KEY_PASSWORD}
```
sscaling: adding the `-storetype=PKCS12` fixes a lot of warnings about the proprietary JKS format. I'm not sure if that fundamentally breaks any of the other security settings

ssl-bootstrap.sh
```diff
@@ -0,0 +1,109 @@
...
+
+if [[ -f /run/secrets/server.truststore.jks ]]; then
+   export KAFKA_SSL_TRUSTSTORE_LOCATION=/run/secrets/server.truststore.jks
+fi
+
+#If the Keystore file already exists (i.e., has been volume-mounted or loaded as a secret) don't run this script any more
+if [[ ! -f ${KAFKA_SSL_KEYSTORE_LOCATION} ]]; then
+
+if [[ -z "$CERT_DN" ]]; then
+export CERT_DN="cn=$(hostname), ou=GenericOU, o=GenericO, c=US"
```
sscaling: $(hostname) will return the internal docker hostname / docker version-dependent hostname - , i.e. with
```
$ openssl s_client -showcerts -connect localhost:9094 </dev/null

... snip ...
Server certificate
subject=/C=US/O=GenericO/OU=GenericOU/CN=b0c33c7326c1
issuer=/C=US/ST=California/L=Cupertino/CN=b0c33c7326c1
```
What is the strategy for accessing broker from outside the docker network?

tcjennings: If relying on autogenerated certificates in this case, the strategy would be to configure the client to ignore certificate issues as in the typical case with self-signed certificates.

kkzz8888:
I am looking to use the this image for SSL secured kafka broker. Looks like it has been done yet? (this is closed as enhancement request. thanks!

Issue #138, opened 2016-11-09
(Error) /tmp/download-kafka.sh: Text file busy


pazams:
An error while building the dockerfile:  /tmp/download-kafka.sh: Text file busy

This is related to https://github.com/docker/docker/issues/9547.

I will PR the fix.

PR #257, opened 2018-01-30
Allow cross building kafka for various archs

sudswas:
This commit introduces a way by which we can now build multi-arch
docker images of kakfa. This patch provides an initial framework
based on some inspiration from:
https://github.com/kubernetes/kubernetes/tree/master/build/debian-base

This patch will allow you to build a cross platform image on an
amd64 machine.

Usage:

     `make build` ==> Retains the old behavior of builds
     `make build ARCH=ppc64le` ==> Builds it for a specific arch

We can further include s390x and arm as other platforms.
Further we can use the manifest-tool or the latest docker client
to push manifest images to dockerhub.

### DIFF
--- a/Dockerfile.crossbuild
+++ b/Dockerfile.crossbuild
@@ -12,11 +12,10 @@ ENV KAFKA_VERSION=$kafka_version \
     PATH=${PATH}:${KAFKA_HOME}/bin
 
 COPY download-kafka.sh start-kafka.sh broker-list.sh create-topics.sh /tmp/
-RUN mkdir -p /opt
 RUN apk add --update unzip wget curl docker jq coreutils bash \
  && chmod a+x /tmp/*.sh && sync \
  && mv /tmp/start-kafka.sh /tmp/broker-list.sh /tmp/create-topics.sh /usr/bin \
- && /tmp/download-kafka.sh \
+ && /tmp/download-kafka.sh && mkdir -p /opt \
  && tar xfz /tmp/kafka_${SCALA_VERSION}-${KAFKA_VERSION}.tgz -C /opt \
  && rm /tmp/kafka_${SCALA_VERSION}-${KAFKA_VERSION}.tgz \
  && ln -s /opt/kafka_${SCALA_VERSION}-${KAFKA_VERSION} /opt/kafka \
--- a/Dockerfile.crossbuild
+++ b/Dockerfile.crossbuild
@@ -1,4 +1,5 @@
-FROM anapsix/alpine-java
+FROM BASEIMAGE
+CROSS_BUILD_COPY qemu-ARCH-static /usr/bin/
 
 ARG kafka_version=1.0.0
 ARG scala_version=2.12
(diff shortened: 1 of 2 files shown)


Dockerfile.crossbuild
```diff
@@ -11,9 +12,9 @@ ENV KAFKA_VERSION=$kafka_version \
     PATH=${PATH}:${KAFKA_HOME}/bin
 
 COPY download-kafka.sh start-kafka.sh broker-list.sh create-topics.sh /tmp/
-
-RUN apk add --update unzip wget curl docker jq coreutils \
- && chmod a+x /tmp/*.sh \
+RUN mkdir -p /opt
```
jaceklaskowski: I think this and the following `RUN`s could benefit from being a single `RUN`, i.e. one layer less.

Dockerfile.crossbuild
```diff
@@ -11,9 +12,9 @@ ENV KAFKA_VERSION=$kafka_version \
...
+RUN apk add --update unzip wget curl docker jq coreutils bash \
+ && chmod a+x /tmp/*.sh && sync \
```
jaceklaskowski: Where does this `sync` come from?

Makefile
```diff
@@ -0,0 +1,53 @@
...
+build: 
+
+ifeq ($(ARCH),amd64)
+	cp ./* $(TEMP_DIR)
+	cat Dockerfile.crossbuild \
+                | sed "s|BASEIMAGE|$(BASEIMAGE)|g" \
+                | sed "/CROSS_BUILD_COPY/d" \
+                > $(TEMP_DIR)/Dockerfile.crossbuild
+	# We just build it using the usual process.
+	docker build -t $(BUILD_IMAGE) -f $(TEMP_DIR)/Dockerfile.crossbuild $(TEMP_DIR)
```
jaceklaskowski: How about replacing `crossbuild` with `ARCH` (since it's no longer crossbuild)?

Dockerfile.crossbuild
sudswas: Will fix in the next patch. Thanks for the comment.

Dockerfile.crossbuild
sudswas: Without this `sync` - I was seeing this issue :  https://github.com/wurstmeister/kafka-docker/issues/138
So i forward ported the fix that's presently here: https://github.com/wurstmeister/kafka-docker/blob/0.10.0/Dockerfile#L12

Makefile
sudswas: Are you suggesting changing the name of the file?
Ideally I kept one file to avoid any out of sync issues across ARCHs. 
Another way could be that I write `> $(TEMP_DIR)/Dockerfile.ARCH` instead of `>$(TEMP_DIR)/Dockerfile.crossbuild` and then build using that file.  Thoughts?

sudswas:
@jaceklaskowski thanks for the quick review. I have updated the changes per your review comments. Let me know your thoughts!

jaceklaskowski:
LGTM

sudswas:
@wurstmeister Hi, could you please review this PR? Thanks!

sscaling:
Hi @sudswas - could you provide some more context to what problem this is solving, along with relevant use-cases and proposed build strategy? Currently images are built on Docker hub, which is relying on the Dockerfile being present.

wurstmeister:
multi arch support added in https://github.com/wurstmeister/kafka-docker/pull/694

PR #267, opened 2018-02-14
fixed path in ENV

AranDekar:
ENV PATH=${PATH}:${KAFKA_HOME}/bin should be in a separate line otherwise it doesn't include ${KAFKA_HOME}/bin


jaceklaskowski:
What do you see in `PATH` when it's together with the other ENVs? Could this be that system env vars are not resolved when chained?

AranDekar:
@jaceklaskowski , that is right, when chained in one command, it cannot resolve it.

sscaling:
@AranDekar - I'm trying to work out if this is actually an issue with this image, or with the docker build process. The Dockerfile docs are not explicit on how ENV interacts with a container instantiated from the image (https://docs.docker.com/engine/reference/builder/#env)

However, in my experience the variables defined in ENV are available to the built container. This is also strange as the CMD definition relies on the PATH environment variable being correct.

Is this actually a Docker/Moby bug?

AranDekar:
@sscaling - I updated to the latest version where all variables were defined in one command (multiple lines) and then found out that cannot run the following command from command prompt:
`<addr>`docker exec -it tss_kafka_1 kafka-topics.sh --list --zookeeper zookeeper:2181
where I used to run it in previous versions, the error was related to unknown path of kafka/bin, I separated the paths in the docker file as committed and then it worked. to see the differences you can open the container and then investigate the PATH like below:
$PATH
and you see that your one is different from this:
/bin/sh: /usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/opt/jdk/bin:/opt/kafka/bin:
the last part (opt/kafka/bin) is missing if you have the env config all in one command.

sscaling:
@AranDekar - I'm not disputing your change provides a work-around for the issue, I just don't think it's the root cause. The official docs don't state any difference in behaviour between the two version of ENV. So I think the root cause is actually on the docker side.

Digging a bit deeper, I think this is the bit of code that munges stuff together [dispatchers.go](https://github.com/moby/moby/blob/master/builder/dockerfile/dispatchers.go#L39) the logic reads to me that only the first variable in a multi env list will get updated

```
for i, envVar := range runConfig.Env {
			envParts := strings.SplitN(envVar, "=", 2)
			compareFrom := envParts[0]
			if shell.EqualEnvKeys(compareFrom, name) {
				runConfig.Env[i] = newVar
				gotOne = true
				break
			}
}
```

So this still feels more like a docker bug. What do you think?

sscaling:
@AranDekar - The code I previously referenced, is actually correct, the issue actually stems from further upstream. 

I had a play around and debugged  the `docker build` code, and found it is the way that `ENV` vars are expanded before being evaluated. So in the current code, `KAFKA_HOME` isn't available because the `ENV` command hasn't been evaluated - hence why the old code (and your proposed) fix works.

Reference, see [https://github.com/moby/moby/blob/master/builder/dockerfile/evaluator.go#L49](https://github.com/moby/moby/blob/master/builder/dockerfile/evaluator.go#L49) **expand**, and later on [https://github.com/moby/moby/blob/master/builder/dockerfile/evaluator.go#L70](https://github.com/moby/moby/blob/master/builder/dockerfile/evaluator.go#L70) **evaluate**. 

There was some refactoring of this code last year (around August 2017) but I have found similar-ish issues from much further back in the main docker repo. So it looks like this is a known issue that they are unlikely to fix/change anytime soon.

Dockerfile
```diff
@@ -7,19 +7,20 @@ MAINTAINER wurstmeister
...
 
 RUN apk add --update unzip wget curl docker jq coreutils \
- && chmod a+x /tmp/*.sh \
- && mv /tmp/start-kafka.sh /tmp/broker-list.sh /tmp/create-topics.sh /usr/bin \
- && /tmp/download-kafka.sh \
- && tar xfz /tmp/kafka_${SCALA_VERSION}-${KAFKA_VERSION}.tgz -C /opt \
- && rm /tmp/kafka_${SCALA_VERSION}-${KAFKA_VERSION}.tgz \
- && ln -s /opt/kafka_${SCALA_VERSION}-${KAFKA_VERSION} /opt/kafka \
- && rm /tmp/*
+    && chmod a+x /tmp/*.sh \
```
sscaling: Can we revert these unrelated whitespace changes

Dockerfile
```diff
@@ -7,19 +7,20 @@ MAINTAINER wurstmeister
 
 ENV KAFKA_VERSION=$kafka_version \
     SCALA_VERSION=$scala_version \
-    KAFKA_HOME=/opt/kafka \
-    PATH=${PATH}:${KAFKA_HOME}/bin
+    KAFKA_HOME=/opt/kafka 
+
+ENV PATH=${PATH}:${KAFKA_HOME}/bin
```
sscaling: With the recent test updates, you should be able to add a test for this now. i.e. add a `test/xxx-test-path.kafka.sh` script that does something like:

```
echo "Checking PATH '$PATH'"
[[ "$PATH" ~= "/opt/kafka/bin" ]]
```

This will help make sure it doesn't get removed in future. Thanks

AranDekar: done

AranDekar:
@sscaling, changes are done, wasn't able to run the test though, I saw the result of 002-... test only.

Dockerfile
sscaling: Sorry, looks like a typo in my suggested ad hoc test code. Should be '=~' test (not the invalid  '~=')

test/Readme.md should give instructions on how to execute tests locally

sscaling:
@AranDekar - not sure if you saw my inline comment. The test that you added is failing because the `~=` assertion should be `=~` (my bad - type in my original suggestion). Then this looks good to merge. Thanks.

AranDekar:
@sscaling - sorry for delay. done.

PR #271, opened 2018-03-06
Add initial travis ci config with a single kafkacat based test

sscaling:
Eventually we might need to switch to something a little easier for more complex tests, such as Java/Go/Python based tests. But kafkacat and bash should be sufficient for initial smoke tests.


wurstmeister:
Thank you!

PR #273, opened 2018-03-07
Explicitly add Zookeeper dependencies for Kafka in the README

magic890:
If you start this image with `docker-compose` and you don't have a running instance of Zookeeper, Kafka will go in exception due to unavailability of Zookeeper terminating the service.

E.g. exception:
```
 FATAL Fatal error during KafkaServer startup. Prepare to shutdown (kafka.server.KafkaServer)
kafka_1       | org.I0Itec.zkclient.exception.ZkTimeoutException: Unable to connect to zookeeper server ''
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,6 +15,7 @@ The image is available directly from [Docker Hub](https://hub.docker.com/r/wurst
 
 - install docker-compose [https://docs.docker.com/compose/install/](https://docs.docker.com/compose/install/)
 - modify the ```KAFKA_ADVERTISED_HOST_NAME``` in ```docker-compose.yml``` to match your docker host IP (Note: Do not use localhost or 127.0.0.1 as the host ip if you want to run multiple brokers.)
+- Zookeeper running instance. It can be retrieved from [wurstmeister/zookeeper](https://hub.docker.com/r/wurstmeister/zookeeper/).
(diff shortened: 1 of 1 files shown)


sscaling:
That's exactly what docker-compose is for, it defines your application. The example [docker-compose.yml](https://github.com/wurstmeister/kafka-docker/blob/master/docker-compose.yml) demonstrates this. Are you seeing issues where Zookeeper is not started?

magic890:
I saw now that file in the repository, using that one or the one that I wrote I don't see any issue.

However I went through the README in this repo and on Docker Hub and I found only partial examples on how to write the `docker-compose` file and I didn't notice that in the root there were some examples.

I suggest to add in the README an explicit sentence to check out the examples and put those in a folder containing all the possible configurations.

Placed in the root of the project they seem source files.

sscaling:
The documentation specifically does not include 'all possible configurations' because the implementation of this Docker image is dynamic. i.e. the same configuration scripts can be used for multiple versions of Kafka.

I agree there could be some merit in moving docker-compose files to an examples/ directory, but the Readme mentions the docker-compose file immediately before and after where you are suggesting your modification. If this is confusing, it would seem perhaps better to clarify which docker-compose file it is referring to. i.e. change it to a link? What do you think?

PR #274, opened 2018-03-07
Support Docker for Mac native

### LINKED ISSUE #110: Not works on Docker Mac Native - Zookeeper timeouts
I'm using Docker Native with 6GB RAM, 2CPU Core.

``` shell
$ docker -v
Docker version 1.12.0, build 8eab29e
```
- `kafka_2.11-0.10.0.0, zookeeper-3.4.6`

I try make steps in [help](http://wurstmeister.github.io/kafka-docker/) for testing.

``` shell
$ docker-compose up -d
$ docker-compose scale kafka=2
$ ./start-kafka-shell.sh 172.18.0.2 172.18.0.3:2181
bash-4.3# $KAFKA_HOME/bin/kafka-topics.sh --create --topic topic \
> --partitions 4 --zookeeper $ZK --replication-factor 2
Exception in thread "main" org.I0Itec.zkclient.exception.ZkTimeoutException: Unable to connect to zookeeper server within timeout: 30000
    at org.I0Itec.zkclient.ZkClient.connect(ZkClient.java:1232)
    at org.I0Itec.zkclient.ZkClient.<init>(ZkClient.java:156)
    at org.I0Itec.zkclient.ZkClient.<init>(ZkClient.java:130)
    at kafka.utils.ZkUtils$.createZkClientAndConnection(ZkUtils.scala:75)
    at kafka.utils.ZkUtils$.apply(ZkUtils.scala:57)
    at kafka.admin.TopicCommand$.main(TopicCommand.scala:54)
    at kafka.admin.TopicCommand.main(TopicCommand.scala)
```

Any idea why this not work?
It's some simple test for zookeeper how check if works correctly?
IP Addresses are from `docker inspect`

sbgreene1307:
Added two files to fix issues with using docker for mac. 

The new compose yaml uses the docker.for.mac.localhost to configure the KAFKA_ADVERTISED_HOST_NAME.
The new start-kafka-shell script accepts a third option to connect to the network generated by the compose yaml. This connects the container, created in the shell script, to the network generated by docker-compose. It allows you to use the docker-compose service names in the start-kafka-shell script instead of the container ip addresses. 

Resolves #110

### DIFF
--- a/README.md
+++ b/README.md
@@ -172,3 +172,12 @@ See the included sample compose file ```docker-compose-swarm.yml```
 ## Tutorial
 
 [http://wurstmeister.github.io/kafka-docker/](http://wurstmeister.github.io/kafka-docker/)
+
+## Docker OSX Native
+
+Use the docker-compose-osx-native.yaml to start your containers. This docker-compose file uses the docker.for.mac.localhost network feature from docker.
+
+To connect to the kafka container: 
+`./start-kafka-shell-osx-native.sh kafka zookeeper:2181 kafkadocker_default`
+
+This script accepts a third argument that to set the docker network to connect to. This allows you to use the named containers from the docker-compose and alleviates the issues with kafka not reaching zookeeper when creating topics with the kafka scripts. 
--- a/docker-compose-osx-native.yaml
+++ b/docker-compose-osx-native.yaml
@@ -1,0 +1,15 @@
+version: '2'
+services:
+  zookeeper:
+    image: wurstmeister/zookeeper
+    ports:
+      - "2181:2181"
+  kafka:
+    build: .
+    ports:
+      - "9092"
+    environment:
+      KAFKA_ADVERTISED_HOST_NAME: docker.for.mac.localhost
+      KAFKA_ZOOKEEPER_CONNECT: zookeeper:2181
+    volumes:
+      - /var/run/docker.sock:/var/run/docker.sock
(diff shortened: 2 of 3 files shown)


elvinx:
+1

sscaling:
I'm not sure this is the best way to solve this issue - I'll take at look at it. I think partly, we need to update some of the deprecated settings, such as advertised.host -> advertised.listeners. But will see if that affects this specific issue.

sscaling:
Thanks for the submission, but this is not a fix, but simply a work-around for mis-configuration.

`docker.for.mac.localhost` is hostname is an internal DNS A record (`exec` into any container)

```
$ docker exec -it e31b579403b4 /bin/sh
root@e31b579403b4:/# dig docker.for.mac.localhost

; <<>> DiG 9.9.5-3-Ubuntu <<>> docker.for.mac.localhost
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 30477
;; flags: qr rd ra; QUERY: 1, ANSWER: 1, AUTHORITY: 0, ADDITIONAL: 0

;; QUESTION SECTION:
;docker.for.mac.localhost.	IN	A

;; ANSWER SECTION:
docker.for.mac.localhost. 0	IN	A	192.168.65.2

;; Query time: 0 msec
;; SERVER: 127.0.0.11#53(127.0.0.11)
;; WHEN: Fri Mar 23 09:18:07 UTC 2018
;; MSG SIZE  rcvd: 58
```

This will break scaling out the brokers as it is not reachable from the host machine.

Passing in the network to the `start-kafka-shell.sh` script doesn't make a lot of sense. A typical use-case is to use a local docker daemon for testing. If this is the case, to simulate production, you generally need  need to connect to the Kafka brokers from an external network. Changing the shell container to use the same network negates this validation.

PR #277, opened 2018-03-09
Introduce shellcheck for validating scripts

sscaling:
* Enable shellcheck for broker-list.sh file, and fix up errors
* Refactor tests to validate broker-list.sh changes


wurstmeister:
Thank you!

Issue #221, opened 2017-09-12
Configuration of listeners and advertised-listeners


SteveWunderbar:
In the current version the config of listeners and advertised-listeners cannot be set directly using environment variables since the settings are overwritten in `kafka-start.sh`. 
It would be much easier (and also more transparent) if one could configure the brokers with the values which are passed as env-variables. 

Effects: With the current version the brokers cannot connect to each other when using multiple ports (INSIDE, OUTSIDE, etc..) because of a broken config. Currently I run all the containers on the same host.

**What I configured with env-variables:**
```KAFKA_ADVERTISED_HOST_NAME: my.host
KAFKA_LISTENER_SECURITY_PROTOCOL_MAP: INSIDE:PLAINTEXT,OUTSIDE:PLAINTEXT
KAFKA_ADVERTISED_PROTOCOL_NAME: OUTSIDE
KAFKA_PROTOCOL_NAME: INSIDE
KAFKA_ADVERTISED_PORT: my_outside_port
KAFKA_PORT: my_inside_port
KAFKA_BROKER_ID: -1
KAFKA_ZOOKEEPER_CONNECT: my.host:zookeeper_port
```

**What the resulting config should look like:**
```
advertised.host.name = null
advertised.listeners = OUTSIDE://my.host:xxxx,INSIDE://my.host:yyyy
advertised.port = null
```
...
```
host.name = 
inter.broker.listener.name = INSIDE
listener.security.protocol.map = INSIDE:PLAINTEXT,OUTSIDE:PLAINTEXT
listeners = OUTSIDE://:xxxx,INSIDE://:yyyy
```
...

**What it actually looks like:**
```
advertised.host.name = null
advertised.listeners = OUTSIDE://my.host:xxxx,INSIDE://my.host:yyyy
advertised.port = null
```
...
```
host.name = 
inter.broker.listener.name = INSIDE
listener.security.protocol.map = INSIDE:PLAINTEXT,OUTSIDE:PLAINTEXT
listeners =  #OUTSIDE://:9096,INSIDE://my.host:yyyy
```
...

I assume the problem with this is that Kafka tries to bind the port `yyyy` on `my.host` twice, which is not possible.

It would be nice to just use the env-variables like this (without any overwriting of `kafka-start.sh`):
```
KAFKA_ADVERTISED_LISTENERS: OUTSIDE://my.host:xxxx,INSIDE://my.host:yyyy
[...]

PR #286, opened 2018-03-18
Remove prescriptive listener configuration

sscaling:
* Update Readme and tests to reflect listener changes
* Add tests for advertised.listeners and listeners
* Restore broker config between each test


sscaling:
Currently advertised.listeners/listeners must be configured in a specific way. To me, it makes more sense for the user to be able to specify the canonical kafka configuration names as per other configuration. This also simplifies the start-kafka script.

It would be good to get feedback from existing users, and other people to test this before merging.

This has the potential to break peoples existing configuration if they are using the prescriptive listener configuration method.

This also fixes #221

README.md
```diff
@@ -78,57 +78,35 @@ HOSTNAME_COMMAND=wget -t3 -T2 -qO-  http://169.254.169.254/latest/meta-data/loca
...
 
-      KAFKA_JMX_OPTS: "-Dcom.sun.management.jmxremote -Dcom.sun.management.jmxremote.authenticate=false -Dcom.sun.management.jmxremote.ssl=false -Djava.rmi.server.hostname=127.0.0.1 -Dcom.sun.management.jmxremote.rmi.port=1099"
-      JMX_PORT: 1099
-
-Jconsole can now connect at ```jconsole 192.168.99.100:1099```
+```
+KAFKA_ADVERTISED_LISTENERS=SSL://_{HOSTNAME_COMMAND}:9093,PLAINTEXT://9092
+```
 
 ## Listener Configuration
```
sscaling: Moved this listener information next to the advertised host for locality of reference.

test/test.start-kafka-advertised-host.kafka.sh
```diff
@@ -6,15 +6,13 @@ testAdvertisedHost() {
...
-	export KAFKA_PORT=2222
 
 	# When the script is invoked
 	source "$START_KAFKA"
 
 	# Then the configuration file is correct
-	assertExpectedConfig 'advertised.listeners=PLAINTEXT://monkey:8888'
-	assertExpectedConfig 'listeners=PLAINTEXT://internal-monkey:2222'
+	assertAbsent 'advertised.listeners'
+	assertExpectedConfig 'listeners=PLAINTEXT://monkey:8888'
```
sscaling: This needs fixing. If the user provides Advertised Host then this should set the advertised.listeners. It will break existing configs using ADVERTISED_HOST_NAME as the brokers wont' be able to resolve 'monkey' in the above case.

sscaling: This needs fixing. If the user provides Advertised Host then this should set the advertised.listeners. It will break existing configs using ADVERTISED_HOST_NAME as the brokers wont' be able to resolve 'monkey' in the above case.

start-kafka.sh
```diff
@@ -35,83 +41,41 @@ if [[ -n "$KAFKA_HEAP_OPTS" ]]; then
...
+        echo "       Please define KAFKA_LISTENERS / (deprecated) KAFKA_ADVERTISED_HOST_NAME"
+        exit 1
+    fi
 
-if [[ -n "$KAFKA_ADVERTISED_LISTENERS" ]]; then
-  unset KAFKA_ADVERTISED_PORT
-  unset KAFKA_ADVERTISED_HOST_NAME
+    # Maintain existing behaviour
+    # If HOSTNAME_COMMAND is provided, set that to the advertised.host.name value if listeners are not defined.
+    export KAFKA_ADVERTISED_HOST_NAME="$HOSTNAME_VALUE"
```
sscaling: Rather than trying to convert everything to 'kafka.listeners', it seems like a better approach is to fall back to what the user has configured (or the documented behaviour of using the `HOSTNAME_COMMAND` value).

This has the added benefit that we should be able to merge this into pre 0.9.0 branch and still work.

Issue #163, opened 2017-02-03
Establish a procedure for backporting changes to master to all published Kafka releases


dead10ck:
There are currently a lot of fixes tangential to the version of Kafka that have been merged into master, but not backported to all the older versions of Kafka. The only difference between any published version of an image from this repo should be the Kafka version, and any operational steps needed to get a particular version working that differ from other versions.

sscaling:
First I think we need to generate deltas between master and versions to see what the drift is.

sscaling:
From some basic analysis: https://gist.github.com/sscaling/96bd7864253a39ab938a4f6649a39c5f - I can't see any reason to not rebuild all supported versions against the `master` code base.

We would have a build matrix such as

| Kafka/Scala | 2.11 | 2.12 |
|-------------|-----|-----|
| 0.8.2.2 | Y | x |
| 0.9.0.1 | Y | x |
| 0.10.2.1 | x | Y |
| 0.11.0.1 | x | Y |
| 1.0.1 | x | Y |
| 1.1.0 | x | Y |

sscaling:
I think the tests will need re-working to target specific versions (for those that are not applicable)

PR #289, opened 2018-03-25
Update kafka 10 to newer alpine

YonatanKiron:
Motivation - current version doesn't use /etc/resolv.conf which prevents
kubernetes usage with the current image.

See https://github.com/gliderlabs/docker-alpine/issues/8

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,11 +2,16 @@ FROM anapsix/alpine-java
 
 MAINTAINER Wurstmeister 
 
-RUN apk add --update unzip wget curl docker jq coreutils
+RUN echo "http://dl-cdn.alpinelinux.org/alpine/v3.7/main" > /etc/apk/repositories && \
+    apk update && \
+    apk add --update unzip wget curl jq coreutils
 
 ENV KAFKA_VERSION="0.10.0.0" SCALA_VERSION="2.11"
 ADD download-kafka.sh /tmp/download-kafka.sh
-RUN /tmp/download-kafka.sh && tar xfz /tmp/kafka_${SCALA_VERSION}-${KAFKA_VERSION}.tgz -C /opt && rm /tmp/kafka_${SCALA_VERSION}-${KAFKA_VERSION}.tgz
+RUN /tmp/download-kafka.sh && \
+    tar xfz /tmp/kafka_${SCALA_VERSION}-${KAFKA_VERSION}.tgz -C /opt && \
+    rm /tmp/kafka_${SCALA_VERSION}-${KAFKA_VERSION}.tgz && \
+    rm -rf /var/cache/apk/*
 
 VOLUME ["/kafka"]
 
--- a/download-kafka.sh
+++ b/download-kafka.sh
@@ -1,5 +1,5 @@
 #!/bin/sh
 
-mirror=$(curl --stderr /dev/null https://www.apache.org/dyn/closer.cgi\?as_json\=1 | jq -r '.preferred')
+mirror="https://archive.apache.org/dist/"
 url="${mirror}kafka/${KAFKA_VERSION}/kafka_${SCALA_VERSION}-${KAFKA_VERSION}.tgz"
 wget -q "${url}" -O "/tmp/kafka_${SCALA_VERSION}-${KAFKA_VERSION}.tgz"


Dockerfile
```diff
@@ -2,11 +2,16 @@ FROM anapsix/alpine-java
...
+    apk update && \
+    apk add --update unzip wget curl jq coreutils
 
 ENV KAFKA_VERSION="0.10.0.0" SCALA_VERSION="2.11"
 ADD download-kafka.sh /tmp/download-kafka.sh
-RUN /tmp/download-kafka.sh && tar xfz /tmp/kafka_${SCALA_VERSION}-${KAFKA_VERSION}.tgz -C /opt && rm /tmp/kafka_${SCALA_VERSION}-${KAFKA_VERSION}.tgz
+RUN /tmp/download-kafka.sh && \
+    tar xfz /tmp/kafka_${SCALA_VERSION}-${KAFKA_VERSION}.tgz -C /opt && \
+    rm /tmp/kafka_${SCALA_VERSION}-${KAFKA_VERSION}.tgz && \
+    rm -rf /var/cache/apk/*
```
sscaling: This is now in a separate layer, so won't save any space.

Dockerfile
```diff
@@ -2,11 +2,16 @@ FROM anapsix/alpine-java
 
 MAINTAINER Wurstmeister 
 
-RUN apk add --update unzip wget curl docker jq coreutils
+RUN echo "http://dl-cdn.alpinelinux.org/alpine/v3.7/main" > /etc/apk/repositories && \
```
sscaling: TBH - we plan to switch to the openjdk alpine image (something like https://github.com/wurstmeister/kafka-docker/pull/290) - so this shouldn't be required.

download-kafka.sh
```diff
@@ -1,5 +1,5 @@
 #!/bin/sh
 
-mirror=$(curl --stderr /dev/null https://www.apache.org/dyn/closer.cgi\?as_json\=1 | jq -r '.preferred')
+mirror="https://archive.apache.org/dist/"
```
sscaling: Why have you removed mirror selection?

Dockerfile
YonatanKiron: @sscaling would you prefer that I will switch to `openjdk alpine image` ?

download-kafka.sh
YonatanKiron: Actually because `0.10.0.0` doesn't exists there (Take a look at http://apache.spd.co.il/kafka/)

sscaling: `closer.cgi` just rotates through the various mirrors closest to you.  So if a particular mirror doesn't have the requisite version of kafka, then retrying should eventually result in a mirror that does have the version.

Dockerfile
sscaling: I'm saying that PR #290 supersedes this PR. So if we merge that, this can be closed.

sscaling: PR #290 is merged. Does this fix your issue?

YonatanKiron: @sscaling Not really if it doesn't build older versions.
See I'm using `wurstmeister/kafka:0.10.0.0`, so if you can rebuild the images somehow so it will actually take effect on the images that would be amazing

download-kafka.sh
YonatanKiron: To be true, it doesn't exists in my region but I see what you are talking about

Dockerfile
sscaling: @YonatanKiron - we're trying to position the current master branch so that we can hopefully unify master and release branches. This issue was raised a while ago: https://github.com/wurstmeister/kafka-docker/issues/163

download-kafka.sh
sscaling: Perhaps there's a better technique / algorithm to employ here? Or just make it user overridable - such as EXPLICIT_KAFKA_DOWNLOAD_MIRROR or similar?

Dockerfile
YonatanKiron: Then, yes I think it will fix my issue!

YonatanKiron:
Closing since #290 and Issue #163 are generalizing this issue to all versions

PR #290, opened 2018-03-26
Switch to openJDK (JRE) base image

sscaling:


cdaringe:
i'm not sure how the dockerhub builds are configured, but would this patch over previous release tags?  i use a slightly older version (currently supported from this project), and would want this goodness ASAP, not only on build-forward images.  thx!

sscaling:
@cdaringe - currently no. There are a number of things in-flight at the moment before we can hopefully start unifying master and version branches. There is an open issue for this: https://github.com/wurstmeister/kafka-docker/issues/163. But this requires a number of other changes such as https://github.com/wurstmeister/kafka-docker/pull/286 for compatibility with kafka < 0.9

Issue #208, opened 2017-08-08
Allow multiline definition of KAFKA_CREATE_TOPICS


stphngrtz:
Hi! It would be nice to be able to specify the topics to be created on startup with newlines.

```yaml
    ...
      KAFKA_CREATE_TOPICS: |
        topic1:100:3
        topic2:100:3
        topic3:100:3
        ...
```
docker-compose.yml

aleksandr-vin:
One can use this workaround:

```
    environment:
       KAFKA_CREATE_TOPICS:
         "topic1:100:3\
          topic2:100:3\
          topic3:100:3"
```

grigorovich-czi:
There is an issue with that. On first run it works well. But if you stop container and then run it again, it will fail with such error
```
kafka_1            | Excluding KAFKA_HOME from broker config
kafka_1            | waiting for kafka to be ready
kafka_1            | [Configuring] 'port' in '/opt/kafka/config/server.properties'
kafka_1            | [Configuring] 'create.topics.separator' in '/opt/kafka/config/server.properties'
kafka_1            | sed: unmatched '@
```
server.properties at that moment looks like
```
...
create.topics.separator=$'
'
```
Not sure, but this single quote on next line probably breaks something.

gustavosoares:
Having the same issue as reported above

```
Starting dev_kafka_1 ... done
Attaching to dev_kafka_1
kafka_1      | Excluding KAFKA_HOME from broker config
kafka_1      | waiting for kafka to be ready
kafka_1      | [Configuring] 'advertised.host.name' in '/opt/kafka/config/server.properties'
kafka_1      | [Configuring] 'port' in '/opt/kafka/config/server.properties'
kafka_1      | [Configuring] 'create.topics.separator' in '/opt/kafka/config/server.properties'
kafka_1      | sed: unmatched '@'
dev_kafka_1 exited with code 1
```

tvhung83:
+1 same issue with `docker-compose` config:
```yaml
    environment:
      KAFKA_ADVERTISED_HOST_NAME: broker
      KAFKA_ZOOKEEPER_CONNECT: zookeeper:2181
      KAFKA_CREATE_TOPICS_SEPARATOR: "$$'\n'"
      KAFKA_CREATE_TOPICS: |
        topic_1:1:1
        topic_2:1:1
```

gerardobort:
[...]

PR #299, opened 2018-04-04
Allow any config parameter in KAFKA_CREATE_TOPICS

towhans:


towhans:
With this fix you can do:

`KAFKA_CREATE_TOPICS: "Topic3:1:1:delete:message.timestamp.type=LogAppendTime"`

which was previously not possible. Only replication factor, number of partitions and cleanup policy could be specified. Now you can add arbitrary config parameter separated by `:` as shown in the example.

create-topics.sh
```diff
@@ -35,7 +35,14 @@ IFS=','; for topicToCreate in $KAFKA_CREATE_TOPICS; do
...
         config="--config=cleanup.policy=${topicConfig[3]}"
     fi
-    JMX_PORT='' "$KAFKA_HOME/bin/kafka-topics.sh" --create --zookeeper "$KAFKA_ZOOKEEPER_CONNECT" --replication-factor "${topicConfig[2]}" --partitions "${topicConfig[1]}" --topic "${topicConfig[0]}" "$config" --if-not-exists &
+    for i in {4..100}
+    do
+        if [ -n "${topicConfig[i]}" ]; then
+            config="${config} --config=${topicConfig[i]}"
+        fi
+    done
+    COMMAND="${KAFKA_HOME}/bin/kafka-topics.sh --create --zookeeper ${KAFKA_ZOOKEEPER_CONNECT} --replication-factor ${topicConfig[2]} --partitions ${topicConfig[1]} --topic ${topicConfig[0]} ${config} --if-not-exists &"
```
jaceklaskowski: While you're at it, why don't you switch `--replication-factor` with `--partitions` so it's consistent with the numbers?

I'd also use `\` to add new lines and make the single command a bit more readable.

create-topics.sh
```diff
@@ -35,7 +35,14 @@ IFS=','; for topicToCreate in $KAFKA_CREATE_TOPICS; do
     if [ -n "${topicConfig[3]}" ]; then
         config="--config=cleanup.policy=${topicConfig[3]}"
     fi
-    JMX_PORT='' "$KAFKA_HOME/bin/kafka-topics.sh" --create --zookeeper "$KAFKA_ZOOKEEPER_CONNECT" --replication-factor "${topicConfig[2]}" --partitions "${topicConfig[1]}" --topic "${topicConfig[0]}" "$config" --if-not-exists &
+    for i in {4..100}
```
sscaling: This seems a little inconsistent now to set topic configuration in two separate ways. Is this really needed?

create-topics.sh
```diff
@@ -35,7 +35,14 @@ IFS=','; for topicToCreate in $KAFKA_CREATE_TOPICS; do
...
+    do
+        if [ -n "${topicConfig[i]}" ]; then
+            config="${config} --config=${topicConfig[i]}"
+        fi
+    done
+    COMMAND="${KAFKA_HOME}/bin/kafka-topics.sh --create --zookeeper ${KAFKA_ZOOKEEPER_CONNECT} --replication-factor ${topicConfig[2]} --partitions ${topicConfig[1]} --topic ${topicConfig[0]} ${config} --if-not-exists &"
+    JMX_PORT='' eval "${COMMAND}"
```
sscaling: Have you tested with JMX_PORT set. I don't think this value will get passed through to the evaluated command.

create-topics.sh
sscaling: This is something that should have a test

create-topics.sh
towhans: I agree. I was thinking about backward compatibility when doing it this way. Removing the first 3 hard coded config keys would be more consistent though.

create-topics.sh
towhans: I did not. I will test it.

sscaling: Thanks

create-topics.sh
sscaling: I think the first 3 make sense, as they are the minimum topic configuration (name / partitions / replication factor). However, the others are all advanced config, so makes sense to call out specifically in key/value format. (i.e. anything that gets passed to `--config`).

I recently added an 'Announcement' section to the README which has all breaking changes in, and there is also a section explaining how `KAFKA_CREATE_TOPICS` works (https://github.com/wurstmeister/kafka-docker#automatically-create-topics) - this should also be updated to reflect any changes. Thanks. 

If you don't have the inclination, I can always tidy up the README/CHANGELOG etc later.

towhans:
As for the need to do this at all - in Erlang/Elixir world the client library does not expose the API to create topics. Only two options left are autocreate and KAFKA_CREATE_TOPICS. AND we need to create topics with different configurations. Also it is a nice way to specify topics in a configuration file.

sscaling:
@towhans - Typically the topic creation is an operational / orchestration task, not something done by the consumer/producer clients. This is exactly what the `kafka-topics.sh` scripts are for. 

I believe the only reason this feature is here, is because kafka 0.8 had bugs in auto-topic creation, so this just allowed people to create topics out of the box so they weren't stumped at the first hurdle.

I'm not saying this is a bad idea, potentially very useful - I was just wondering what the demand was to fully configure topics in this way vs. the traditional orchestration methods.

sscaling:
@towhans - I think we should have consistent `--config` parameters. It seems a small breaking change to have users update their config from: 

```
foo:1:2:compact
```

to

```
foo:1:2:cleanup.policy=compact
```

Considering they'd now be able to override all config parameters.

Have you considered all the various inputs to topic config: https://kafka.apache.org/documentation/#topicconfigs

Currently your change will break with anything that accepts a list (conflicting comma), currently `follower.replication.throttled.replicas` and `leader.replication.throttled.replicas`. 

A separate YAML is a possibility, but then we have to parse and extract values which I think will require extra tooling in the image / flakey bash parsing. This also requires the user to mount / re-build image with their file for this to work.

It's also worth noting, that all these topic policies can be configured as defaults in the broker. The only use-case this serves is for differing policies across multiple topics on the same brokers.

sscaling:
One final thought on this, that I had last night, is that the user can currently configure all options simply by leveraging the string interpolation, i.e. 
```
KAFKA_CREATE_TOPICS=topic:1:2:compact --config=compression.type=snappy
```
This has the same limitations of conflicting commas that separate each topic definition.

There is another issue open #208 for supporting multiline topic creation commands. (I put together a PR for this https://github.com/wurstmeister/kafka-docker/pull/305)

This made me start to wonder if it's  just better to add a new command such as `CREATE_TOPICS_ARGS` to be a more generic multiline definition of input into the `kafka-topics.sh` script, i.e.

```
CREATE_TOPICS_ARGS: |
  --topic foo --partitions 2 --replication-factor 1 --config cleanup.policy=compact --config compression.type=snappy
  --topic bar --partitions 1 --replication-factor 1
```

create-topics.sh
```diff
@@ -28,14 +28,11 @@ fi
 
 # Expected format:
 #   name:partitions:replicas:cleanup.policy:<key>=<value>:<key>=<value>...
-IFS=','; for topicToCreate in $KAFKA_CREATE_TOPICS; do
+IFS=';'; for topicToCreate in $KAFKA_CREATE_TOPICS; do
```
jaceklaskowski: I think that was just changed in another PR today and so you should wait till it's approved.

towhans: OK, I will wait and rebase

towhans:
Yes, the multiline topic creation is much prettier. 

I tied the:

`KAFKA_CREATE_TOPICS=topic:1:2:compact --config=compression.type=snappy`

but it did not work for me. Bash translated it as it `--config 'compact --config=compression.type=snappy'`

create-topics.sh
```diff
@@ -27,15 +27,26 @@ if $start_timeout_exceeded; then
...
-    fi
-    JMX_PORT='' "$KAFKA_HOME/bin/kafka-topics.sh" --create --zookeeper "$KAFKA_ZOOKEEPER_CONNECT" --replication-factor "${topicConfig[2]}" --partitions "${topicConfig[1]}" --topic "${topicConfig[0]}" "$config" --if-not-exists &
+    config=""
+    for i in {3..100}
+    do
+        if [ -n "${topicConfig[i]}" ]; then
+            config="${config} --config=${topicConfig[i]}"
+        fi
+    done
+    COMMAND="JMX_PORT='' ${KAFKA_HOME}/bin/kafka-topics.sh
```
sscaling: I think if you move to the eval approach, then my earlier suggestion of 

```
topic:1:2:compact --config=compression.type=snappy
```

Works fine - negating the need for this to be a breaking change.

However, this should still have a test-case added to verify this behaviour

create-topics.sh
```diff
@@ -27,15 +27,26 @@ if $start_timeout_exceeded; then
...
+		--create
+		--zookeeper ${KAFKA_ZOOKEEPER_CONNECT}
+		--topic ${topicConfig[0]}
+		--partitions ${topicConfig[1]}
+		--replication-factor ${topicConfig[2]}
+		${config}
+		--if-not-exists &"
+    eval "${COMMAND//$'\n'/}"
```
sscaling: I think it's better just to suffix a `\` to each line in the conventional bash multi-line way, then you can just eval the string, rather than assigning it to `COMMAND` first.

create-topics.sh
```diff
@@ -27,15 +27,26 @@ if $start_timeout_exceeded; then
 fi
 
 # Expected format:
-#   name:partitions:replicas:cleanup.policy
-IFS=','; for topicToCreate in $KAFKA_CREATE_TOPICS; do
+#   name:partitions:replicas:cleanup.policy:<key>=<value>:<key>=<value>...
+IFS=';'; for topicToCreate in $KAFKA_CREATE_TOPICS; do
```
sscaling: Not needed since https://github.com/wurstmeister/kafka-docker/pull/305, so this can be made as a non-breaking change.

create-topics.sh
```diff
@@ -27,15 +27,26 @@ if $start_timeout_exceeded; then
...
     echo "creating topics: $topicToCreate"
     IFS=':' read -r -a topicConfig <<< "$topicToCreate"
-    config=
-    if [ -n "${topicConfig[3]}" ]; then
-        config="--config=cleanup.policy=${topicConfig[3]}"
-    fi
-    JMX_PORT='' "$KAFKA_HOME/bin/kafka-topics.sh" --create --zookeeper "$KAFKA_ZOOKEEPER_CONNECT" --replication-factor "${topicConfig[2]}" --partitions "${topicConfig[1]}" --topic "${topicConfig[0]}" "$config" --if-not-exists &
+    config=""
```
sscaling: can revert

test/test.create-topics.kafka.sh
```diff
@@ -13,7 +13,7 @@ testCreateTopics() {
 	TOPICS[1]="compact-$NOW"
 	CLEANUP[1]="compact"
 
-	KAFKA_CREATE_TOPICS="${TOPICS[0]}:1:1,${TOPICS[1]}:2:1:compact" create-topics.sh
+	KAFKA_CREATE_TOPICS="${TOPICS[0]}:1:1;${TOPICS[1]}:2:1:cleanup.policy=compact" create-topics.sh
```
sscaling: can revert.

README.md
```diff
@@ -15,6 +15,7 @@ The image is available directly from [Docker Hub](https://hub.docker.com/r/wurst
 
 ## Announcements
 
+* **06-Apr-2018** - *BREAKING* - `KAFKA_CREATE_TOPICS` has different format
```
sscaling: can revert

README.md
```diff
@@ -68,7 +69,7 @@ added in ```docker-compose.yml```.
 Here is an example snippet from ```docker-compose.yml```:
 
         environment:
-          KAFKA_CREATE_TOPICS: "Topic1:1:3,Topic2:1:1:compact"
+          KAFKA_CREATE_TOPICS: "Topic1:1:3;Topic2:1:1:cleanup.policy=compact"
```
sscaling: can revert

create-topics.sh
```diff
@@ -27,15 +27,26 @@ if $start_timeout_exceeded; then
 fi
 
 # Expected format:
-#   name:partitions:replicas:cleanup.policy
-IFS=','; for topicToCreate in $KAFKA_CREATE_TOPICS; do
+#   name:partitions:replicas:cleanup.policy:<key>=<value>:<key>=<value>...
```
sscaling: can be reverted

towhans:
Made all the reverts and changes. Works as:

```
CREATE_TOPICS_ARGS: |
   topic:1:2:compact --config=compression.type=snappy
   topic:2:1:compact
```

How do you run the tests? I get:

```
T560:~/git/kafka-docker[master] > test/runAllTests.sh 
test/runAllTests.sh: line 3: ./internal-broker-list.sh: No such file or directory
```

PR #301, opened 2018-04-04
Update KAFKA_LISTENERS example with correct syntax

samling:
Updating KAFKA_LISTENERS example on line 129 to have a colon in front of the port number, i.e. `INSIDE://9092` -> `INSIDE://:9092`. Container dies with a fatal error otherwise.


sscaling:
Thanks!

Issue #303, opened 2018-04-05
1.0.1 version does not support snappy compession


gburanov:
Hi. 

I saw you recently (in this commit https://github.com/wurstmeister/kafka-docker/commit/7a25ade4cc820ec9e5a103a0b83f412008d3351e#diff-3254677a7917c6c01f55212f86c57fbf) introduced new base image

The problem with the new base image is that it lacks libc.

https://github.com/xerial/snappy-java (that is used by kafka) uses libc and fails without it

This means that snappy compression is broken now =(
Can we add extra step to install libc?

sscaling:
Duplicates https://github.com/wurstmeister/kafka-docker/issues/298

sscaling:
Also worth noting that 1.0.1 was cut before the change to the openjdk alpine base image. So if you simple use the wurstmeister/kafka:1.0.1 build this should address any immediate issue.

PR #304, opened 2018-04-05
Add libc to image

gburanov:
I described here why do I actually need libc
https://github.com/wurstmeister/kafka-docker/issues/303


sscaling:
Duplicates https://github.com/wurstmeister/kafka-docker/pull/306

Thanks for the PR, however as this is actually currently broken functionality I've added the working code and supporting test in #306.

PR #306, opened 2018-04-08
[BUG] include glibc for SNAPPY support

### LINKED ISSUE #298: Error using snappy compression
I believe the cause is a missing glibc library since the base image is using Alpine. Kafka cannot load the snappy compression library which has a dependency on it.

Exception will look like this in Kafka:

```java
java.lang.NoClassDefFoundError: Could not initialize class org.xerial.snappy.Snappy
at org.xerial.snappy.SnappyInputStream.hasNextChunk(SnappyInputStream.java:435)
at org.xerial.snappy.SnappyInputStream.read(SnappyInputStream.java:466)
at java.io.DataInputStream.readByte(DataInputStream.java:265)
```

To reproduce add to Kafka with a producer that has snappy compression enabled
```java
config.put(ProducerConfig.COMPRESSION_TYPE_CONFIG, "snappy");
```

See also: https://github.com/xerial/snappy-java/issues/181

sscaling:
Fixes #298

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,10 +3,11 @@ Changelog
 
 Kafka features are not tied to a specific kafka-docker version (ideally all changes will be merged into all branches). Therefore, this changelog will track changes to the image by date.
 
-TBC
----
+08-Apr-2018
+-----------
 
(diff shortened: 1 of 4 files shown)


sscaling:
Uses same method of installation as the old anapsix/alpine-java image: https://github.com/anapsix/docker-alpine-java/blob/master/Dockerfile.server-jre.tpl

PR #308, opened 2018-04-09
docker-compose-swarm.yml has the wrong container name for kafka

dsevilla:
It read ${REPO}kafka-docker instead of ${REPO}kafka

### DIFF
--- a/docker-compose-swarm.yml
+++ b/docker-compose-swarm.yml
@@ -5,7 +5,7 @@ services:
     ports:
       - "2181:2181"
   kafka:
-    image: kafka:latest
+    image: wurstmeister/kafka:latest
     ports:
       - target: 9094
         published: 9094
--- a/docker-compose-swarm.yml
+++ b/docker-compose-swarm.yml
@@ -5,7 +5,7 @@ services:
     ports:
       - "2181:2181"
   kafka:
-    image: ${REPO}kafka-docker:latest
+    image: ${REPO}kafka:latest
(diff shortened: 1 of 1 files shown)


docker-compose-swarm.yml
```diff
@@ -5,7 +5,7 @@ services:
     ports:
       - "2181:2181"
   kafka:
-    image: ${REPO}kafka-docker:latest
+    image: ${REPO}kafka:latest
```
sscaling: i'm not sure why this even uses the `${REPO}` environment var. It should just be the canonical `wurstmeister/kafka:latest` image

dsevilla: I think it's there as a leftover of some tests using docker stacks. In stacks you usually build a temporal registry, and you have to tag the images something like hostname:port/kafka (e.g. 127.0.0.1:5000/kafka). Anyway, once the image is stable and available in the docker hub, no need for this REPO variable, so it should be removed in the github distribution.

docker-compose-swarm.yml
```diff
@@ -5,7 +5,7 @@ services:
     ports:
       - "2181:2181"
   kafka:
-    image: ${REPO}kafka-docker:latest
+    image: kafka:latest
```
sscaling: This should be `wurstmeister/kafka:latest` (so it references the docker hub build)

sscaling:
Thanks!

Issue #312, opened 2018-04-11
SASL_PLAIN/SSL KAFKA_OPTS not being set


jduffy3:
I'm trying to pass some variables into the KAFKA_OPTS var but 
I keep seeing this in the logs

/usr/bin/start-kafka.sh: line 96: KAFKA_OPTS=-Djava.security.auth.login.config: bad substitution

I'm trying to set the following variable
KAFKA_OPTS: "-Djava.security.auth.login.config=/kafka_server_jaas.conf"

Can you please explain why I can't set this? Or how I might go about doing it

sscaling:
I believe his is part of a larger namespace collision problem that affects multiple elements such as Kubernetes deployments etc (as well as other KAFKA_ service settings).

Given you are referencing an external file `/kafka_server_jaas.conf`, i'm assuming you're OK adding/mounting extra files through; a work-around is to specify a `CUSTOM_INIT_SCRIPT` environment var, which should be a script similar to:

```
#!/bin/bash
export KAFKA_OPTS="-Djava.security.auth.login.config=/kafka_server_jaas.conf"
```

This is executed after the substitution part that is failing.

This could have been done inline, however there is currently a bug in how we process the environment, where we need to specify the input separator to make this work correctly.

sscaling:
Have opened https://github.com/wurstmeister/kafka-docker/issues/313 to track the second issue I mentioned.

jduffy3:
I tried the following but it hasn't corrected it. 
```
Is there something I'm missing?

  kafka:
    image: 'wurstmeister/kafka:latest'
    volumes:
      - "./env.sh:/usr/local/bin/env.sh"
    ports:
      - '9092:9092'
      - '9093:9093'
    environment:
      KAFKA_ZOOKEEPER_CONNECT: zookeeper:2181
      KAFKA_HOST_NAME: 10.10.10.1
      KAFKA_PORT: 9092
      KAFKA_ADVERTISED_PORT: 9093
      KAFKA_ADVERTISED_HOST_NAME: 10.10.10.1
      KAFKA_LISTENERS: PLAINTEXT://:9092,SASL_SSL://:9093
      KAFKA_ADVERTISED_LISTENERS: PLAINTEXT://10.10.10.1:9092,SASL_SSL://10.10.10.1:9093
      KAFKA_SECURITY_INTER_BROKER_PROTOCOL: SASL_SSL
      KAFKA_SASL_ENABLED_MECHANISMS: PLAIN
[...]

PR #317, opened 2018-04-17
Fix regex when parsing env vars

Morlinest:
Fix regex, otherwise script ends with:

> ... -Dcom.sun.management.jmxremote.rmi.port: bad substitution

Also unifying variable names.

### DIFF
--- a/start-kafka.sh
+++ b/start-kafka.sh
@@ -99,18 +99,18 @@ echo "" >> "$KAFKA_HOME/config/server.properties"
     for VAR in $(env)
     do
       if [[ $VAR =~ ^KAFKA_ && ! $VAR =~ ^KAFKA_HOME ]]; then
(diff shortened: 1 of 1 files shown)


sscaling:
Relates to: https://github.com/wurstmeister/kafka-docker/issues/312

As per the above bug, there is a work around for this issue using `CUSTOM_INIT_SCRIPT`. However, the actual issue here is that we're processing un-related variables and trying to add them to the server.properties file,  i.e. we should never even parse the `KAFKA_OPTS` var (or whatever you are using) because we do not want to inject it into the broker config.

I think to address #312 - we actually need to use a blacklist to exclude any envs that we know about that should not be injected (KAFKA_OPTS, KAFKA_HEAP_OPTS etc). This change just stops the error occurring. Ultimately, it would be better to not use the conflicting `KAFKA_` namespace, but changing that will most likely break every config out there :(

sscaling:
I've prepared the other part of the fix for #312, so will merge this now, rebase this and submit a PR for the other changes.

sscaling:
Thanks!

Issue #310, opened 2018-04-10
download-kafka.sh sometimes fails


sscaling:
Currently the `download-kafka.sh` script can fail depending on what region you are in.

It uses the following apache `closer.cgi` script to find the closest geographical mirror.  However, we don't tell it what resource we are looking for, so this can potentially lead to finding a mirror without the required resource.

We use

```
https://www.apache.org/dyn/closer.cgi?as_json=1
```

But we should add the following query parameter

```
path=/kafka/${KAFKA_VERSION}/kafka_${SCALA_VERSION}-${KAFKA_VERSION}.tgz'
```

e.g.

```
https://www.apache.org/dyn/closer.cgi?path=/kafka/0.8.2.1/kafka_2.11-0.8.2.1.tgz&as_json=1'
```

PR #318, opened 2018-04-17
extend download-kafka.sh script

mielmat:
I made changes in download-kafka.sh script according to issue description, any comments are welcome

### DIFF
--- a/download-kafka.sh
+++ b/download-kafka.sh
@@ -1,7 +1,7 @@
 #!/bin/sh -e
 
 path="kafka/${KAFKA_VERSION}/kafka_${SCALA_VERSION}-${KAFKA_VERSION}.tgz"
-mirror=$(curl --stderr /dev/null 'https://www.apache.org/dyn/closer.cgi?path=/${path}&as_json=1' | jq -r '.preferred')
+mirror=$(curl --stderr /dev/null "https://www.apache.org/dyn/closer.cgi?path=/${path}&as_json=1" | jq -r '.preferred')
 
 if [[ -z "$mirror" ]]; then
 	echo "Unable to determine mirror for downloading Kafka, the service may be down"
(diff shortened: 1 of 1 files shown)


download-kafka.sh
```diff
@@ -1,11 +1,12 @@
...
+path="kafka/${KAFKA_VERSION}/kafka_${SCALA_VERSION}-${KAFKA_VERSION}.tgz"
+mirror=$(curl --stderr /dev/null "https://www.apache.org/dyn/closer.cgi?path=/${path}&as_json=1" | jq -r '.preferred')
 
 if [[ -z "$mirror" ]]; then
 	echo "Unable to determine mirror for downloading Kafka, the service may be down"
 	exit 1
 fi
 
-url="${mirror}kafka/${KAFKA_VERSION}/kafka_${SCALA_VERSION}-${KAFKA_VERSION}.tgz"
+url=${mirror}${path}
```
sscaling: This looks a bit weird from a bash scripting perspective. Perhaps it's better to just use the response from `closer.cgi` to build the full URL, such as:

```
url=$(curl "https://www.apache.org/dyn/closer.cgi?path=${path}&as_json=1" | jq -r '"\(.preferred)\(.path_info)")'
```

mielmat: Thanks for review, I made changes in my pull request

download-kafka.sh
```diff
@@ -1,11 +1,12 @@
 #!/bin/sh -e
+KAFKA_VERSION=1.1.0
```
sscaling: KAFKA_VERSION and SCALA_VERSION are provided by the Dockerfile so shouldn’t be added here.

sscaling:
Thanks!

sscaling:
Fixes https://github.com/wurstmeister/kafka-docker/issues/310

PR #328, opened 2018-05-12
Provide override mechanism for versions pre-bootstrap

sscaling:
* Remove listeners for 0.9.0.1 kafka config, so bootstrap process can be the same


sscaling:
Part of support work for https://github.com/wurstmeister/kafka-docker/issues/163

Issue #311, opened 2018-04-10
Document extended create topic options


sscaling:
See details in PR #299

PR #333, opened 2018-05-23
Add KAFKA_CONFIGS variable to do additional configuration

raarts:
I needed more specific per-topic settings.
This PR allows using the kafka-configs.sh script during creation time.


sscaling:
This is already supported. Please see example in https://github.com/wurstmeister/kafka-docker/blob/master/test/0.10/test.create-topics.kafka.sh#L16

raarts:
I see the test, but I can't explain how it works, where is the extra --config parameter parsed in create-config.sh (looking at https://github.com/wurstmeister/kafka-docker/blob/master/create-topics.sh)

Does kafka-topics.sh accept multiple --config arguments? I remember trying and it did not work.

sscaling:
Yes - admittedly it's not obvious, but implemented to support backwards compatibility.

```
test:2:1:compact --config=<x> --config=<y> --config=<z>
```

Where `compact` is topic compaction policy. Documentation of this is an existing issue: https://github.com/wurstmeister/kafka-docker/issues/311

sscaling:
> ok, I think I understand now. Should be documented better.

happy to close this PR?

sscaling:
Thanks!

PR #335, opened 2018-05-29
Build docker images from travis-ci

### LINKED ISSUE #163: Establish a procedure for backporting changes to master to all published Kafka releases

### LINKED ISSUE #210: Link between 'latest' tag and Kafka version
Hi,

We want to integrate this docker image in our build. However, we have to ship and maintain multiple versions of our software. Therefore, we want to have a good understanding of which version of Kafka is running in each of our release version, and currently, we don't understand which version of kafka corresponds to the image tagged 'latest'. 
My understanding is that you tag the latest release of the docker image as 'latest' (and versioned all the older ones) but you don't provide an equivalent image tagged with the corresponding version.  
Would it be possible to have two images, one tagged 'latest' and one tagged with the kafka version?

Thanks for your consideration.

Louis

### LINKED ISSUE #231: keytool is missing from versions 0.9
The `keytool` seems to be missing from the 0.9 images.

```fish
~/s/r/k/tests (test ± ) ∫ docker run --rm wurstmeister/kafka:0.8.2.2 find / -name keytool
/etc/alternatives/keytool
/var/lib/dpkg/alternatives/keytool
/usr/bin/keytool
/usr/lib/jvm/java-6-openjdk-amd64/bin/keytool
/usr/lib/jvm/java-6-openjdk-amd64/jre/bin/keytool
~/s/r/k/tests (test ± ) ∫ docker run --rm wurstmeister/kafka:0.9.0.0 find / -name keytool
~/s/r/k/tests (test ± ) ∫ docker run --rm wurstmeister/kafka:0.9.0.1 find / -name keytool
~/s/r/k/tests (test ± ) ∫ docker run --rm wurstmeister/kafka:0.10.2.1 find / -name keytool
/opt/jdk1.8.0_131/jre/bin/keytool
~/s/r/k/tests (test ± ) ∫ docker run --rm wurstmeister/kafka:latest find / -name keytool
/opt/jdk1.8.0_144/jre/bin/keytool
```

#163 would *really* help with things like this. The base image was switched to alpine in 0.9, and clearly the 0.8 images haven't been touched since then.

### LINKED ISSUE #324: KAFKA_JMX_OPTS causing start-kafka.sh to fail (Issue reopen)
I'm getting this error on kafka logs:

```/usr/bin/start-kafka.sh: line 107: KAFKA_JMX_OPTS=-Dcom.sun.management.jmxremote -Dcom.sun.management.jmxremote.authenticate=false -Dcom.sun.management.jmxremote.ssl=false -Djava.rmi.server.hostname=127.0.0.1 -Dcom.sun.management.jmxremote.rmi.port: bad substitution```

When KAFKA_JMX_OPTS is defined as:
`KAFKA_JMX_OPTS: "-Dcom.sun.management.jmxremote -Dcom.sun.management.jmxremote.authenticate=false -Dcom.sun.management.jmxremote.ssl=false -Djava.rmi.server.hostname=127.0.0.1 -Dcom.sun.management.jmxremote.rmi.port=9010"`

This error happens only when upgrading to version 1.1.0 (on version 1.0.1 it doesn't happen).

### PR DESCRIPTION
**Pull request by sscaling**
* Fix up downloading binaries < 0.11
* Update kafkacat tests to account for older broker versions

Fixes #163
Fixes #324
Fixes #210
Fixes #231
Fixes #210

**TODO**

- [x] Configure travis-ci with dockerhub user/pass
- [ ] Decide on if we should only build from `release/` branch

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,14 +1,52 @@
 sudo: required
 
-language: java
+language: scala
 
 services:
   - docker
 
+# An alternative approach is to build a full matrix.
+#scala:
+#  - "2.10"
+#  - 2.11
+#
+#env:
+#  - KAFKA_VERSION=0.8.2.2
+#  - KAFKA_VERSION=0.9.0.1
+#  - KAFKA_VERSION=0.10.2.1
+#  - KAFKA_VERSION=0.11.0.2
+#  - KAFKA_VERSION=1.0.1
+#  - KAFKA_VERSION=1.1.0
+
+# This version will be also tagged as 'latest'
+env:
+  global:
+    - LATEST="2.11-1.1.0"
+
+# Build recommended versions based on: http://kafka.apache.org/downloads
+matrix:
+  include:
+  - scala: "2.10"
+    env: KAFKA_VERSION=0.8.2.2
+  - scala: 2.11
+    env: KAFKA_VERSION=0.9.0.1
+  - scala: 2.11
+    env: KAFKA_VERSION=0.10.2.1
+  - scala: 2.11
+    env: KAFKA_VERSION=0.11.0.2
+  - scala: 2.11
+    env: KAFKA_VERSION=1.0.1
+  - scala: 2.11
+    env: KAFKA_VERSION=1.1.0
+
 install:
   - docker --version
   - docker-compose --version
-  - docker build -t wurstmeister/kafka .
+  - echo "KAFKA VERSION  $KAFKA_VERSION"
+  - echo "SCALA VERSION  $TRAVIS_SCALA_VERSION"
+  - echo "LATEST VERSION $LATEST"
+  - export CURRENT=${TRAVIS_SCALA_VERSION}-${KAFKA_VERSION}
+  - docker build --build-arg kafka_version=$KAFKA_VERSION --build-arg scala_version=$TRAVIS_SCALA_VERSION -t wurstmeister/kafka .
   - docker pull confluentinc/cp-kafkacat
 
 before_script:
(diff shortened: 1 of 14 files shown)

### CODE REVIEW

File: .travis.yml [hunk 1]
```diff
@@ -1,14 +1,52 @@
...
+#  - KAFKA_VERSION=0.11.0.2
+#  - KAFKA_VERSION=1.0.1
+#  - KAFKA_VERSION=1.1.0
+
+# This version will be also tagged as 'latest'
+env:
+  global:
+    - LATEST="2.11-1.1.0"
+
+# Build recommended versions based on: http://kafka.apache.org/downloads
```
sscaling: While getting familiar with this setup and to keep build times relatively low and not use up a lot of travis-ci resources, I've stuck to the main major versions of each release along with the recommended scala version.

If we go down the official image route, then we should probably support different architectures as well - so I thought this should also be considered before we end up building 100's of images with every change :)

.travis.yml
```diff
@@ -1,14 +1,52 @@
...
+#  - KAFKA_VERSION=0.9.0.1
+#  - KAFKA_VERSION=0.10.2.1
+#  - KAFKA_VERSION=0.11.0.2
+#  - KAFKA_VERSION=1.0.1
+#  - KAFKA_VERSION=1.1.0
+
+# This version will be also tagged as 'latest'
+env:
+  global:
+    - LATEST="2.11-1.1.0"
```
sscaling: This could also be parsed from the main Dockerfile to keep this info in one place. However, I thought this change was big enough and could see pros/cons to both approaches.

matlockx:
👍

PR #336, opened 2018-06-01
describe KAFKA_ZOOKEEPER_CONNECT value's format

vitaliis:
couple of days ago I tried to use the latest image, but faced a problem starting the container
Neither README.md nor web-page described "KAFKA_ZOOKEEPER_CONNECT" value.
There was no documentation how to use it and what should be the value for this variable.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,14 +16,14 @@ The image is available directly from [Docker Hub](https://hub.docker.com/r/wurst
 ## Announcements
 
 * **03-Apr-2018** - *BREAKING* - `KAFKA_ADVERTISED_PROTOCOL_NAME` and `KAFKA_PROTOCOL_NAME` removed. Please update to canonical kafka settings.
-* **03-Apr-2018** - *BREAKING* - `KAFKA_ZOOKEEPER_CONNECT` is now a mandatory environment var (see examples).
+* **03-Apr-2018** - *BREAKING* - `KAFKA_ZOOKEEPER_CONNECT` is now a mandatory environment var.
 
 ---
 
 ## Pre-Requisites
 
 - install docker-compose [https://docs.docker.com/compose/install/](https://docs.docker.com/compose/install/)
-- modify the ```KAFKA_ADVERTISED_HOST_NAME``` in ```docker-compose.yml``` to match your docker host IP (Note: Do not use localhost or 127.0.0.1 as the host ip if you want to run multiple brokers.)
+- modify the ```KAFKA_ADVERTISED_HOST_NAME``` in [docker-compose.yml](https://raw.githubusercontent.com/wurstmeister/kafka-docker/master/docker-compose.yml) to match your docker host IP (Note: Do not use localhost or 127.0.0.1 as the host ip if you want to run multiple brokers.)
 - if you want to customize any Kafka parameters, simply add them as environment variables in ```docker-compose.yml```, e.g. in order to increase the ```message.max.bytes``` parameter set the environment to ```KAFKA_MESSAGE_MAX_BYTES: 2000000```. To turn off automatic topic creation set ```KAFKA_AUTO_CREATE_TOPICS_ENABLE: 'false'```
 - Kafka's log4j usage can be customized by adding environment variables prefixed with ```LOG4J_```. These will be mapped to ```log4j.properties```. For example: ```LOG4J_LOGGER_KAFKA_AUTHORIZER_LOGGER=DEBUG, authorizerAppender```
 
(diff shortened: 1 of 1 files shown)


sscaling:
All the example docker compose files use this:

* https://github.com/wurstmeister/kafka-docker/blob/master/docker-compose-single-broker.yml
* https://github.com/wurstmeister/kafka-docker/blob/master/docker-compose-swarm.yml
* https://github.com/wurstmeister/kafka-docker/blob/master/docker-compose.yml
* https://github.com/wurstmeister/kafka-docker/blob/master/test/docker-compose.yml
* https://github.com/wurstmeister/kafka-docker/blob/master/test/scenarios/jmx/docker-compose.yml

The README & CHANGELOG document this as a required property.

The Connectivity guide uses this in all examples: https://github.com/wurstmeister/kafka-docker/wiki/Connectivity

What you are proposing is not a pre-requisite for running the example `docker-compose.yml`- but more information for hand rolling a custom configuration (which requires understanding how to configure Kafka).

Perhaps this would be better reworked as a quick-start steps with a link to the example `docker-compose.yml` with a separate section explain about how to specify Kafka settings and required configurations etc (Although the error message should be clear in the logs if you have omitted the zookeeper connect setting). Would this clear up the confusion?

vitaliis:
> All the example docker compose files use this:

Yes, but I haven't seen where these files are referenced. So when I first faced this problem, I could not find a solution and gave up till the next attempt.


.
> The README & CHANGELOG document this as a required property.

Exactly. And at the same time it doesn't provide you any examples of how to specify this required value.



.
> What you are proposing is not a pre-requisite for running the example docker-compose.yml- but more information for hand rolling a custom configuration (which requires understanding how to configure Kafka).

Hm, I thought without it the Kafka wouldn't run and indeed I get `..._kafka_1 exited with code 1`. So it looks like a required parameter now.


I already solved my issue. I just found it unclear from the doc. 
Moreover I haven't found anything except that it is required, from 2 main sources I found:
https://github.com/wurstmeister/kafka-docker
http://wurstmeister.github.io/kafka-docker/

For me it is fine either way. I just want to make documentation a bit more clear for newcomers and enable them to use the solution you provided.

sscaling:
Out of interest, did you come to this image from github or dockerhub ?

vitaliis:
From dockerhub.

Previously we used `image: spotify/kafka` and when I tried to upgrade to the new version I realized there is a single tag called `latest` and no tags by versions. Then I searched there and came across https://hub.docker.com/r/wurstmeister/kafka/tags/ where you have a lot of versions available. And decided to give it a try :)

vitaliis:
Alternatively I can propose to open a new section "Appendix" where this setting as well as examples can be referenced. Although examples, containing links to docker-compose ymls that you mentioned in your first comment, there can be a separate sections.

sscaling:
Hmm, I think that's where the confusion comes from then - as dockerhub doesn't contain all the other files (such as the docker-compose file).

I think really all that should be required is changing the `docker-compose.yml` reference in the pre-requisites section to a link, then the reference is there, i.e.

```
* modify the `KAFKA_ADVERTISED_HOST_NAME` in [docker-compose.yml](https://github.com/wurstmeister/kafka-docker/blob/master/docker-compose.yml) to match your docker host IP (Note: Do not use localhost or 127.0.0.1 as the host ip if you want to run multiple brokers.)
```

Ultimately I think the readme needs a bit of re-working, and providing a quick-start quide for unix / mac / windows etc at the top, and arrange the rest of the information a little better - but I think updating the static text to a link should help avoid people having similar confusion in future?

vitaliis:
But from the user's perspective:
when I found an image I want to start using it with minimal effort. At best having instructions on the dockerhub or in README. Next maybe on the related web site (http://wurstmeister.github.io/kafka-docker/). And the last thing I want to do is to browse some source code in order to find the missing peace that is needed to start using the thing.
Other option I see: provide a complete example in the README (e.g. inlining the docker-compose.yml).

sscaling:
If the link is to raw content (i.e. `https://raw.githubusercontent.com/wurstmeister/kafka-docker/master/docker-compose.yml`) then the action is right-click, save-as to get the docker-compose file.

If there is an example to copy & paste, the user has to correctly copy, paste somewhere (if into a text editor this could break the sensitive YAML formatting), and then run.

From experience of sharing YAML, I personally think the first is less error-prone.

However, I think we need to refocus some parts of the README (such as having a quick start guide at the top etc). To aide other users in any potential confusion, I think updating the link is sufficient - unless you want to tackle the broader changes?

Something along the lines of:

----

## Quick-start

### Mac OS X - Docker for mac

```
mkdir kafka-docker && cd $_ && curl -o docker-compose.yml https://raw.githubusercontent.com/wurstmeister/kafka-docker/master/docker-compose.yml && KAFKA_ADVERTISED_HOST_NAME="$(docker-machine ip)" docker-compose up -d
```

### Mac OS X - Docker machine

... 

### Linux

...

### Windows

...

----

vitaliis:
Ok, we can put links to examples. At least users will know where to look.
And the quick start guide is a separate issue and should be tackled in a separate issue/PR.
I'll prepare the change to refer to examples.

sscaling:
Thanks.

I removed the example links - as they are likely to change and it's easier to maintain the single link. I agree that the README can be improved, however - because of the connectivity requirements of Kafka, it's not as easy as providing a few instructions as there are many different sets and varies between mac / linux / windows / swarm / kubernetes / ECS etc. 

Hopefully we can keep making things easier and easier for people to use, but at this stage the user still requires some knowledge of Kafka and Docker to successfully configure the image beyond a single node broker (and even that has caveats)

vitaliis:
But your change has nothing to do with `KAFKA_ZOOKEEPER_CONNECT`.

It is still unclear what is the value format for `KAFKA_ZOOKEEPER_CONNECT`. One of course can guess, and say it is hostname+port separated by the colon. But it is not obvious. The link to overall example is somewhere far away and is not related to this setting.

If you don't want to add the description to "Prerequisites" section, what about adding it right after announcement of this breaking change? E.g.: 
 **03-Apr-2018** - *BREAKING* - `KAFKA_ZOOKEEPER_CONNECT` is now a mandatory environment var. + """Value's format is `host:port`, e.g. `zookeeper:2181`"""

vitaliis:
Ok, maybe it was just me having this confusion... Let's see if this question pops up again in the future ;)

PR #343, opened 2018-06-18
Fix Invalid interpolation and KAFKA_CREATE_TOPICS_SEPARATOR

xtuc:
```
ERROR: Invalid interpolation format for "environment" option in service "kafka": "hostname | awk -F'-' '{print $2}'"
```


sscaling:
Thanks

PR #361, opened 2018-07-23
Update kafka to 1.1.1

jrpilat:
I think I did this right, tests passed locally.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,7 @@ services:
 #  - KAFKA_VERSION=0.10.2.1
 #  - KAFKA_VERSION=0.11.0.2
 #  - KAFKA_VERSION=1.0.1
-#  - KAFKA_VERSION=1.1.0
+#  - KAFKA_VERSION=1.1.1
 
 # This version will be also tagged as 'latest'
 env:
--- a/.travis.yml
+++ b/.travis.yml
@@ -37,8 +37,6 @@ matrix:
   - scala: 2.11
     env: KAFKA_VERSION=1.0.1
   - scala: 2.11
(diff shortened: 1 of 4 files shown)


.travis.yml
```diff
@@ -38,6 +38,8 @@ matrix:
     env: KAFKA_VERSION=1.0.1
   - scala: 2.11
     env: KAFKA_VERSION=1.1.0
+  - scala: 2.11
```
sscaling: This should just replace `1.1.0`. `1.1.1` is the latest version of the `1.1.x branch`.

sscaling:
Thanks!

Dockerfile
```diff
@@ -1,6 +1,6 @@
 FROM openjdk:8u171-jre-alpine
 
-ARG kafka_version=1.1.0
+ARG kafka_version=1.1.1
```
jaceklaskowski: If this Dockerfile uses Kafka 1.1.1 with Scala 2.12 (which is recommended) why does `.travis.yml` use 2.11? Just curious.

sscaling: http://kafka.apache.org/downloads recommends 2.11 with version 1.1.1. Is 2.12 recommended use documented anywhere else? 

Travis ci always passes build arguments, so the default values should probably be removed as they are redundant.

jaceklaskowski: Sorry for the "noise". I thought that 2.12 was the recommended version.

sscaling: no problem. Good to have another pair of eyes. And it highlights the fact that we should probably remove the default values. Will address as part of https://github.com/wurstmeister/kafka-docker/issues/363

PR #369, opened 2018-07-30
Update for Kafka 2.0.0

pwoody:
Mirrors #361 - ran tests locally and it seemed fine. Also ran it against some integration tests for another project and seems to work as well.


sscaling:
Thanks.

PR #371, opened 2018-07-31
Add changing image to Pre-Requisites

Shpota:
It is impossible to start the cluster locally as the docker-compose file from the documentation contains instruction to build the Kafka image instead of using the one from docker hub.

I updated the readme to include one more step to Pre-Requisites.

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,6 +40,7 @@ Everytime the image is updated, all tags will be pushed with the latest updates.
 
 - install docker-compose [https://docs.docker.com/compose/install/](https://docs.docker.com/compose/install/)
 - modify the ```KAFKA_ADVERTISED_HOST_NAME``` in [docker-compose.yml](https://raw.githubusercontent.com/wurstmeister/kafka-docker/master/docker-compose.yml) to match your docker host IP (Note: Do not use localhost or 127.0.0.1 as the host ip if you want to run multiple brokers.)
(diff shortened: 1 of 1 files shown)


sscaling:
The docker compose files are examples and the build command should build the image from source. If you want to use a specific label from docker hub then it's expected you'd need to create your own custom docker-compose file.

1. git clone https://github.com/wurstmeister/kafka-docker.git
2. cd kafka-docker
3. docker-compose up

Is that not working for you?

Shpota:
Your approach will probably work if I want to customize the image, but I just want to use it.

The concern is primarily with [instructions on docker hub](https://hub.docker.com/r/wurstmeister/kafka/). I followed the guide but it simply didn't work. How could one know that it is required to clone the repository prior to running kafka locally? And why would somebody want to build the image instead of using the prebuilt one?

sscaling:
The documentation on dockerhub is no longer updated since we switched to travis-ci for releasing, so any changes here will not be reflected in dockerhub. Unfortunately they don't provide an API to programatically update it. (I've just opened https://github.com/wurstmeister/kafka-docker/issues/373 to manually update it and link to the github README)

If this is really causing a headache, a better solution might be to update the docker-compose file with the latest image defined, rather than using the `build` command. If doing this, we should also update the single broker template. One negative of doing this is that if the image has previously being run, then the latest image will not be pulled unless a docker / docker-compose pull is run first - so we should update the README to reflect this.

PR #383, opened 2018-08-15
Update docker-compose.yml

nabinkhadka:
Now the kafka server will get pinged from outside the server.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -7,7 +7,7 @@ services:
   kafka:
     build: .
     ports:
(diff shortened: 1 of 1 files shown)


sscaling:
Please see https://github.com/wurstmeister/kafka-docker/blob/master/docker-compose-single-broker.yml / https://github.com/wurstmeister/kafka-docker/wiki/Connectivity. This docker-compose file allows scaling the number of brokers on a physical host for testing / experimentation etc. So it is like this by design.

Issue #293, opened 2018-03-29
Update Dockerfile to use labels


sscaling:
See: https://docs.docker.com/config/labels-custom-metadata/ and  http://label-schema.org/rc1/

`MAINTAINER` has been deprecated (https://docs.docker.com/engine/reference/builder/#maintainer-deprecated)

https://github.com/wurstmeister/kafka-docker/pull/256/files#diff-3254677a7917c6c01f55212f86c57fbf is probably a good starting point.

PR #408, opened 2018-10-06
Update Dockerfile to use labels and add vcs ref

ttapjinda:
#293


Dockerfile
```diff
@@ -3,8 +3,16 @@ FROM openjdk:8u171-jre-alpine
...
 ARG glibc_version=2.27-r0
-
-MAINTAINER wurstmeister
+ARG vcs_ref=unspecified
+
+LABEL org.label-schema.name="kafka" \
+      org.label-schema.description="Apache Kafka" \
+      org.label-schema.vcs-url="https://github.com/wurstmeister/kafka-docker" \
+      org.label-schema.vcs-ref="${vcs_ref}" \
+      org.label-schema.schema-version="1.0" \
```
sscaling: build-date seems like another useful addition to add here.

Dockerfile
```diff
@@ -3,8 +3,16 @@ FROM openjdk:8u171-jre-alpine
...
+      maintainer="wurstmeister" \
+      org.apache.kafka.scala-version="${scala_version}" \
```
sscaling: Given that this is not an offiically supported apache.org image, it might be better to use the label-schema suggested label of `version` and remove the `org.apache.xxx` labels i.e.

```
org.label-schema.version="${scala_version}_${kafka_version}"
```

This also matches the tag of the image build on dockerhub.

docker_build.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash -e
```
sscaling: What is this file for - it doesn't seem to be used anywhere?

.travis.yml
```diff
@@ -35,7 +35,7 @@ install:
   - echo "SCALA VERSION  $TRAVIS_SCALA_VERSION"
   - echo "LATEST VERSION $LATEST"
   - export CURRENT=${TRAVIS_SCALA_VERSION}-${KAFKA_VERSION}
-  - docker build --build-arg kafka_version=$KAFKA_VERSION --build-arg scala_version=$TRAVIS_SCALA_VERSION -t wurstmeister/kafka .
+  - docker build --build-arg kafka_version=$KAFKA_VERSION --build-arg scala_version=$TRAVIS_SCALA_VERSION --build-arg vcs_ref=$(git rev-parse --short HEAD) -t wurstmeister/kafka .
```
sscaling: I think this is probably available from travis-ci env (https://docs.travis-ci.com/user/environment-variables/#default-environment-variables)

test/verifyImageLabels.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash -e
```
sscaling: I think this file can be just moved into the generic `test/0.0` folder and run for all builds.

test/verifyImageLabels.sh
```diff
@@ -0,0 +1,9 @@
...
+
+VCS_REF=$(docker inspect -f '{{ index .Config.Labels "org.label-schema.vcs-ref"}}' wurstmeister/kafka)
+echo "VCS_REF=$VCS_REF"
+if [ -z "$VCS_REF" ] || [ "$VCS_REF" = "unspecified" ]; then
```
sscaling: This test could be improved to check that the sha is a valid commit in the master git tree.

ttapjinda:
@sscaling Thank you for reviewing, your points sound valid to me. I'll update the code.

test/verifyImageLabels.sh
ttapjinda: I put it here because currently all tests under test/[version] are running inside docker ("kafka" or "cp-kafkacat" image). 
but this test's verifying the labels of kafka image itself so I think that the labels should be checked outside.
Also by calling from script in .travis.yml, this should already run for all builds.

test/verifyImageLabels.sh
```diff
@@ -0,0 +1,20 @@
...
+  exit 1
+fi
+if ! git cat-file -e "$VCS_REF^{commit}"; then
+  echo "$VCS_REF Not a valid git commit"
+  exit 1
+fi
+
+BUILD_DATE=$(docker inspect -f '{{ index .Config.Labels "org.label-schema.build-date"}}' wurstmeister/kafka)
+echo "BUILD_DATE=$BUILD_DATE"
+if ! date -d "$BUILD_DATE"; then
```
sscaling: This will fail on MacOS because of the BSD version of date. However, I think this is OK for now - but ideally should be run from inside a container like the other tests so that it can be executed from any machine.

test/verifyImageLabels.sh
sscaling: As per the comment below, it could just be added as another container (using the mounted docker socket) and perform these checks in a consistent way so that anybody can run the tests. However, for now I think this is fine. I'll fix up at a later date if it becomes an issue. However, agreed that it is different from the functionality tests that are currently containerized.

sscaling:
Thanks for the PR and fixing up the comments!

Issue #413, opened 2018-10-22
Use Docker secrets with sensitive environment variables


zeemba:
Hello,

Can we have provisions for setting sensitive environment variables such as KAFKA_SSL_KEYSTORE_PASSWORD, KAFKA_SSL_KEY_PASSWORD and KAFKA_SSL_TRUSTSTORE_PASSWORD to also use docker secrets by adding the "_FILE" suffix

These environment variables could be changed to KAFKA_SSL_KEYSTORE_PASSWORD_FILE, KAFKA_SSL_KEY_PASSWORD_FILE and KAFKA_SSL_TRUSTSTORE_PASSWORD_FILE which basically runs a "cat" command on the location of the secrets in the container "/run/secrets/secret_name".

This would help in removing the plaintext password in the deployment yml file.

Thanks

zeemba:
Oh I had created a PR #415 before your reply.

I tried using the $CUSTOM_INIT_SCRIPT initially by exporting the passwords but it didn't work.

Something like:

`#!/bin/bash
export KAFKA_SSL_KEYSTORE_PASSWORD='password'
`
So I decided to add an extra if statement to the start-kafka.sh file.  I make use of the $CUSTOM_INIT_SCRIPT currently to pass in the KAFKA_OPTS environment variable but I'll also try out your recommendation and get back to you.

Thanks

sscaling:
the `CUSTOM_INIT_SCRIPT` won't work like that for export for at least two reasons: 

1) it is executed via `eval`.
2) it is executed after the `server.properties` interpolation so the environment variables would never be used.

zeemba:
I'm also trying to pass a jmx agent config file with the KAFKA_OPTS environment variable in my $CUSTOM_INIT_SCRIPT which breaks the kafka service when I add the ssl passwords along with it.

`CUSTOM_INIT_SCRIPT: "export KAFKA_OPTS=-javaagent:/usr/app/jmx_prometheus_javaagent.jar=7071:/usr/app/prom-jmx-agent-config.yml"`

Note that the above environment variable works fine when I set my ssl key passwords in plaintext as we currently do.

I also took the KAFKA_OPTS bit out and only did a `cat` on the secrets as recommended. The logs showed the secrets in plain text (which can be changed with the log levels so that's fine) but the kafkaConfig values in the log still ended showing the `ssl.keystore.password = null`  with an error saying 

[...]

PR #415, opened 2018-10-23
Make provisions for passing passwords into the container with Docker secret 

zeemba:
I'm sure there's a better way to do this but I added an extra if statement to cater for passing environment variables like KAFKA_SSL_KEYSTORE_PASSWORD as KAFKA_SSL_KEYSTORE_PASSWORD_FILE instead.

The if statement then copies the content of the file into the $KAFKA_HOME/config/server.properties file.

For example;
environment:
KAFKA_SSL_KEYSTORE_PASSWORD: 'secret'

becomes

environment:
KAFKA_SSL_KEYSTORE_PASSWORD_FILE: '/run/secrets/kafka_keystore_pswd'
secrets:

kafka_keystore_pswd
Then the content of the /run/secrets/kafka_keystore_pswd gets copied into the server.properties file.

I noticed however that an extra property (ssl.keystore.password.file in this case) also gets copied into the server.properties file which isn't really needed but works fine anyway.

### DIFF
--- a/start-kafka.sh
+++ b/start-kafka.sh
@@ -129,6 +129,12 @@ echo "" >> "$KAFKA_HOME/config/server.properties"
             echo "Excluding $env_var from broker config"
             continue
         fi
+	
+	#if the environment variable contains the word "FILE" remove the file part of the environment variable and copy the content of the file into the server.properties file
+	if [[ $env_var == *_FILE* ]]; then
+	    kafka_name=$(echo "$env_var" | cut -d_ -f2- | tr '[:upper:]' '[:lower:]' | tr _ . | rev | cut -c6- | rev)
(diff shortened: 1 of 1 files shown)


sscaling:
We probably want to avoid overloading the `KAFKA_xxx` variables as they are supposed to map directly onto the canonical [kafka](http://kafka.apache.org/documentation.html#brokerconfigs) configuration. As mentioned in the #413 this currently can be solved with a `CUSTOM_INIT_SCRIPT`.

sscaling:
Closing as not required at this point in time - see bug ticket for alternatives. 

Thanks for the PR.

zeemba:
Thanks for the assistance @sscaling

PR #428, opened 2018-11-21
update to newest kafka version 2.1.0

kyleboyle:


sscaling:
Looks good. Just missing the update to the Dockerfile (This information was missing from the release process) so have updated it. I pushed an update to this branch so we can merge it when the build passes.

sscaling:
Thanks

kyleboyle:
Thanks for the quick turnaround.

PR #453, opened 2019-01-25
Minor documentation update

dpimkin:
Documentation update for recent environment variables and shell scripts.


sscaling:
The documentation looks correct to me based on the instructions given. `start-kafka-shell.sh` will configure the `ZK` environment var. See: https://github.com/wurstmeister/kafka-docker/blob/master/start-kafka-shell.sh#L2

PR #458, opened 2019-02-06
Update Dockerfile

### LINKED ISSUE #457: Dockerfile Problem '/opt' directory exists
In Dockerfile there is a problem with RUN command. 
A `/opt` directory exists. This causes the block of composition on a 12 layer. 
Removing `mkdir /opt` command, the `docker-compose up` ends without problem.

angeloparchitelli:
`/opt` folder exists.

Fixes #457


sscaling:
Thanks!

PR #467, opened 2019-02-27
Document HOSTNAME_COMMAND strategy

AndyGee:


README.md
```diff
@@ -110,6 +110,19 @@ HOSTNAME_COMMAND=wget -t3 -T2 -qO-  http://169.254.169.254/latest/meta-data/loca
 ```
 Reference: http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-instance-metadata.html
 
+For Local, OpenShift, Kubernetes or Swarm deployments you can use the following to get the container host's IP:
+```
+HOSTNAME_COMMAND=ip -o -4 addr list eth0 | awk '{print \$4}' | cut -d/ -f1
```
sscaling: This doesn't look like it provides the 'container host's IP' to me. As far as I can tell this returns the internal IP assigned to the container within the assigned docker network (not normally route-able directly from the host)

README.md
```diff
@@ -110,6 +110,19 @@ HOSTNAME_COMMAND=wget -t3 -T2 -qO-  http://169.254.169.254/latest/meta-data/loca
...
+```
+Note: Here `$4` is escaped with `\`, which is dependent on where you use/define the environment variable.
```
sscaling: Typically I have seen people configuring docker containers via docker-compose when there are multiple parameters. Here the escaping only applies to starting things from a `sh` / `bash` shell so isn't universally applicable.

README.md
AndyGee: The internal IP is available/routable to other containers within the same network. This can be useful as the kafka advertized IP/hostname that is provided to zookeeper and kafka clients that obtain the kafka hosts/IPs dynamically. That aside, I've found this useful for testing, where docker-compose is used to start zookeeper and kafka, and the test is started in it's own process.

README.md
AndyGee: That's the point I was trying to make, that it only needs to be escaped on the command line (I should re-word it). I actually found a better one that works for me that doesn't crumble:

`ip -o -4 addr show dev eth0 | cut -d' ' -f7 | cut -d'/' -f1`

README.md
sscaling: 1) The comment is ambiguous. I interpret "container host's IP" as the IP-of-the-host where the container is running. I think you just mean container IP?
2) From your comment, this particular configuration is useful to your use-case but it's not a universal approach - or one suitable if for example you wanted to have access outside of the docker-network.
3) I think by default, the image will do what you want and advertise the docker assigned hostname. Specifying `listeners` of `PLAINTEXT://:9092` should also give you what you want I believe.

README.md
sscaling: There's no example usage or mention of using the command line to manually `docker run` the image. The rest of the documentation assumes/describes docker-compose so this feels out of place.

PR #469, opened 2019-03-01
docs: add quick start, use host.docker.internal for docker host ip

jk2K:
fix: docker-compose.yml version to 3

docker-compose [scale](https://docs.docker.com/compose/reference/scale/) command is deprecated, Use the [up](https://docs.docker.com/compose/reference/up/) command with the --scale flag instead

use [host.docker.internal](host.docker.internal) which resolves to the internal  IP address used by the host

add kafka quick start, refer to https://kafka.apache.org/quickstart

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,8 +39,7 @@ Everytime the image is updated, all tags will be pushed with the latest updates.
 
 ## Pre-Requisites
 
-- install docker-compose [https://docs.docker.com/compose/install/](https://docs.docker.com/compose/install/)
-- modify the ```KAFKA_ADVERTISED_HOST_NAME``` in [docker-compose.yml](https://raw.githubusercontent.com/wurstmeister/kafka-docker/master/docker-compose.yml) to match your docker host IP (Note: Do not use localhost or 127.0.0.1 as the host ip if you want to run multiple brokers.)
+- install docker-compose https://docs.docker.com/compose/install/
 - if you want to customize any Kafka parameters, simply add them as environment variables in ```docker-compose.yml```, e.g. in order to increase the ```message.max.bytes``` parameter set the environment to ```KAFKA_MESSAGE_MAX_BYTES: 2000000```. To turn off automatic topic creation set ```KAFKA_AUTO_CREATE_TOPICS_ENABLE: 'false'```
 - Kafka's log4j usage can be customized by adding environment variables prefixed with ```LOG4J_```. These will be mapped to ```log4j.properties```. For example: ```LOG4J_LOGGER_KAFKA_AUTHORIZER_LOGGER=DEBUG, authorizerAppender```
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -50,15 +49,40 @@ Everytime the image is updated, all tags will be pushed with the latest updates.
...
+```
 
 Destroy a cluster:
 
-- ```docker-compose stop```
+```bash
+docker-compose stop
+```
+
+quick start in kafka container, refer to https://kafka.apache.org/quickstart
```
sscaling: The tutorial already covers what to do.

README.md
```diff
@@ -50,15 +49,40 @@ Everytime the image is updated, all tags will be pushed with the latest updates.
...
-- ```docker-compose up -d ```
+```bash
+docker-compose up -d
+```
 
 Add more brokers:
 
-- ```docker-compose scale kafka=3```
+```bash
+docker-compose up -d --scale kafka=3
```
sscaling: This introduces an implicit dependency on newer versions of docker-compose.

docker-compose.yml
```diff
@@ -9,7 +9,7 @@ services:
     ports:
       - "9092"
     environment:
-      KAFKA_ADVERTISED_HOST_NAME: 192.168.99.100
+      KAFKA_ADVERTISED_HOST_NAME: host.docker.internal
```
sscaling: This breaks external access to the broker and I think is only supported by default for docker for windows.

README.md
```diff
@@ -39,8 +39,7 @@ Everytime the image is updated, all tags will be pushed with the latest updates.
 
 ## Pre-Requisites
 
-- install docker-compose [https://docs.docker.com/compose/install/](https://docs.docker.com/compose/install/)
-- modify the ```KAFKA_ADVERTISED_HOST_NAME``` in [docker-compose.yml](https://raw.githubusercontent.com/wurstmeister/kafka-docker/master/docker-compose.yml) to match your docker host IP (Note: Do not use localhost or 127.0.0.1 as the host ip if you want to run multiple brokers.)
```
sscaling: This is the single most important configuration option, should not be removed.

README.md
```diff
@@ -39,8 +39,7 @@ Everytime the image is updated, all tags will be pushed with the latest updates.
 
 ## Pre-Requisites
 
-- install docker-compose [https://docs.docker.com/compose/install/](https://docs.docker.com/compose/install/)
```
sscaling: The link should be a markdown link. It just so happens that github is rendering the url as a link, other markdown viewers may not.

docker-compose.yml
```diff
@@ -1,4 +1,4 @@
-version: '2'
+version: '3'
```
sscaling: not necessary.

sscaling:
Hi thanks for the PR. However, I can't see any need for these changes, most of which assume that it will be run on a given system or version of Docker.

docker-compose.yml
jk2K: docker for macOS is ok

sscaling: A quick check shows it returns an internal to docker host IP (assuming there's some bridge magic to route the traffic back to the host). However, the existing example shows a docker-machine IP, which is routable from the host->container. I believe `host.docker.internal` is for container->host mapping. This is not useful in this scenario, you might as well just leave blank for container hostname. Please correct me if my understanding is incorrect.

README.md
jk2K: maybe we should use newer versions of docker-compose

docker-compose.yml
jk2K: maybe your are right, I can run multiple brokers when using `host.docker.internal`,  nslookup can get IP addresses like 192.168.65.2, but `host.docker.internal` not work for yahoo/kafka-manager, yahoo/kafka-manager deployed on the docker host

PR #474, opened 2019-03-28
Add a HEALTHCHECK 

### LINKED ISSUE #167: Docker compose health check
docker-compose 1.10.0 now supports health checks, and delaying startup of containers until their dependencies are up and healthy. See https://docs.docker.com/compose/startup-order/ for docs on this.

It would be great with an example of how to configure such health checks on a kafka container!

timdorr:
Fixes #167 

I swiped the script from @knordstrom and @Lukkie. The general gist is that it asks Zookeeper for the brokers it knows about and checks that it is in that list. So, a little better than just a port check with netcat.

If anyone has ideas for improvement, please let me know and I'll add it to this PR.

### DIFF
--- a/healthcheck.sh
+++ b/healthcheck.sh
@@ -1,0 +1,20 @@
+#! /bin/bash
+
+r=`$KAFKA_HOME/bin/zookeeper-shell.sh zookeeper:2181 <<< "ls /brokers/ids" | tail -1 | jq '.[]'`   
+ids=( $r )                                                                                         
+function contains() {
+     local n=$#
+     local value=${!n}
+     for ((i=1;i < $#;i++)) {
+         if [ "${!i}" == "${value}" ]; then
+             echo "y"
+             return 0
+         fi
+     }
+     echo "n"
+     return 1
+}
+
+LOG_DIR=$(awk -F= -v x="log.dirs" '$1==x{print $2}' /opt/kafka/config/server.properties)
+x=`cat ${LOG_DIR}/meta.properties | awk 'BEGIN{FS="="}/^broker.id=/{print $2}'`
+if [ $(contains "${ids[@]}" "$x") == "y" ]; then echo "ok"; exit 0; else echo "doh"; exit 1; fi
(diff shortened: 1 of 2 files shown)


sscaling:
The response in my comment to this issues still applies -  [https://github.com/wurstmeister/kafka-docker/issues/167#issuecomment-377179871](https://github.com/wurstmeister/kafka-docker/issues/167#issuecomment-377179871)

Health checks don't really provide much use other than an a queryable label on the running container.

It'd be great to understand what real-world use-case this particular solution solves to know on whether to proceed or not.

As an aside - i'm not sure this is a great approach to health check, all it does is validate that the broker has registered and been assigned a znode - it has no real indication of whether Kafka is healthy or not, the only half decent way of doing so is via a successful query to the brokers. Even with that said, if your brokers are in a distinct network from your clients and communication is via an external port a health check from inside the broker network becomes misleading, i.e. it could appear healthy but be completely unusable.

fvigotti:
In my installation which may be slightly different but probably not so much..  the 
shell commands ( ie : zookeeper-shell.sh  ) use same env variables as kafka.. 
with the provided script there could be problems ie: if the pod is limited in memory to something less than 200% of the -Xmx or if the kafka is exporting JMX maybe would be usefull to reset/reimport from another name those variables 
```
KAFKA_HEAP_OPTS="-Xms200M -Xmx500M"
## avoid conflicts on jmx ports with kafka 
KAFKA_JMX_OPTS=""
JMX_PORT=""
```

healthcheck.sh
```diff
@@ -0,0 +1,20 @@
+#! /bin/bash
+
+r=`$KAFKA_HOME/bin/zookeeper-shell.sh $KAFKA_ZOOKEEPER_CONNECT <<< "ls /brokers/ids" | tail -1 | jq '.[]'`   
+ids=( $r )
```
jayvdb, line 4: trailing whitespace on this line and the previous line

also fwiw, using "health☑️" as a branch name makes it difficult for others to use your branch.

jayvdb (changes requested):

Issue #407, opened 2018-09-29
Update tutorial


sscaling:
Here is a master ticket to track all the required tutorial updates, rather than having multiple tickets open tracking separate bits of information

**To update**

- [ ] Identify that `KAFKA_ADVERTISED_HOST` must be routable from host / bridge network. (https://github.com/wurstmeister/kafka-docker/issues/110)
- [ ] Link to connectivity guide
- [ ] Document `--from-beginning` console parameter (https://github.com/wurstmeister/kafka-docker/issues/219)
- [ ] Document `--bootstrap-server` option for new consumers (https://github.com/wurstmeister/kafka-docker/issues/243)
- [ ] Enumerate difference between docker-machine + docker-for-mac
- [ ] Highlight cloning the repository first
- [ ] document any differences in versions for the `scale` / `--scale` commands

FinnFrotscher:
Tipp: 
`$  $KAFKA_HOME/bin/kafka-console-consumer.sh --topic=topic --zookeeper=$ZK`
This command throws 2 errors. `--zookeeper` is not recognized and `--bootstrap-server` is missing.
This line works for me:
`bin/kafka-console-consumer.sh --bootstrap-server `broker-list.sh` --topic topic --from-beginning`

barrer:
For security, do not mount the /var/run/docker.sock configuration file into the container.

https://github.com/wurstmeister/kafka-docker/issues/403#issuecomment-448555550

sscaling:
Now that markdown -> site generation has been fixed by https://github.com/wurstmeister/kafka-docker/pull/454, these updates should be possible.

dzmitry-kankalovich:
Follow up on https://github.com/wurstmeister/kafka-docker/issues/110 for MacOS.

Until you explicitly add `--net kafka-docker_default` to the `start-kafka-shell.sh` script, the internal docker DNS resolution for host `zookeeper` won't work, so further examples about producer/consumer cannot be run.

The correct script should look like this:
```
#!/bin/bash
docker run --rm -v /var/run/docker.sock:/var/run/docker.sock --net kafka-docker_default -e HOST_IP=$1 -e ZK=$2 -i -t wurstmeister/kafka /bin/bash
```

PR #481, opened 2019-04-05
Make the need to clone the repository a bit more explicit.

fungiboletus:
I tried by just downloading the `docker-compose.yml` file and it obviously failed.

### DIFF
--- a/index.md
+++ b/index.md
@@ -7,11 +7,12 @@ Cluster Setup
 
 1.	Install [Docker](https://www.docker.io/gettingstarted/#h_installation)
 2.	Install [Compose](http://docs.docker.com/compose/install/)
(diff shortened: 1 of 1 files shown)


sscaling:
Relates to https://github.com/wurstmeister/kafka-docker/issues/407

sscaling:
Essentially I'm a bit torn by this, it's an outstanding task, and it does fix one particular issue. However, if you look at the related #407 ticket you can see that the ways to connect to Kafka via the console consumer/producer have changed over the versions. So part of me thinks that the actual fix is to provide specific example docker-compose (sans the `build` directive) for download and use in tutorials.

With that said, probably no harm in merging this.

PR #493, opened 2019-05-10
add swarm examle where kafka survive restart

gerassimos:
Add a docker-compose example where kafka is configured to use consistent broker ID, hostname and logs dir, so that can survive restart. 

Some additional requirement:
 - Kafka is accessible only from the overlay virtual network of Docker `net-broker`.
 - Kafka is NOT be accessible from the Docker host IP for security reasons

### DIFF
--- a/docker-compose-swarm-2.yml
+++ b/docker-compose-swarm-2.yml
@@ -1,0 +1,46 @@
+version: '3.7'
+services:
+  zookeeper:
+    image: wurstmeister/zookeeper
+    ports:
+      - "2181:2181"
+  
+  kafka:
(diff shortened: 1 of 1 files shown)


sscaling:
This configuration seems unnecessary and/or confusing. As you are proposing not to connect on the `OUTSIDE` defined listener, there is no need to separate them and provide a custom security protocol map. Just define one `PLAINTEXT://kafka:9092` listener definition. However, this is the method used in the [official documentation](https://kafka.apache.org/documentation/#quickstart_multibroker) so I don't think this example provides any new / additional info. Thanks.

Issue #537, opened 2019-10-26
Travis build broken with docker-compose 1.23


sscaling:
Travis is currently running docker-compose 1.23.1 which mangles the container names (appends a random sha to each one). This breaks various parts of the test process, blocking any further changes. 

This change was reverted in 1.23.2 (https://github.com/docker/compose/releases/tag/1.23.2). In the mean time we can be a bit more explicit in the tests to hopefully prevent this kind of breakage in the future.

PR #538, opened 2019-10-26
pin docker compose version

wurstmeister:
see https://github.com/docker/compose/issues/6316#issuecomment-442666873 

fixes https://github.com/wurstmeister/kafka-docker/issues/537

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,6 +9,13 @@ services:
 env:
   global:
     - LATEST="2.12-2.3.0"
+    - DOCKER_COMPOSE_VERSION=1.23.2
+
+before_install:
+  - sudo rm /usr/local/bin/docker-compose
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -9,6 +9,13 @@ services:
 env:
   global:
     - LATEST="2.12-2.3.0"
+    - DOCKER_COMPOSE_VERSION=1.23.2
```
sscaling: Should probably indicate required version in the README / where to find this info. We should probably update info in the Dependencies section of the [ReleaseProcess](https://github.com/wurstmeister/kafka-docker/wiki/ReleaseProcess) wiki as well. 

Also, to me, it feels a little weird to pin docker-compose and not docker version. It seems that we either say this works with a specific environment and be explicit or we don't pin anything and work with "latest" (as defined by travis ci)

PR #539, opened 2019-10-26
Fix compatibility with docker-compose 1.23.1

### LINKED ISSUE #537: Travis build broken with docker-compose 1.23

sscaling:
Fixes #537

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -45,8 +45,7 @@ install:
   - docker pull confluentinc/cp-kafkacat
 
 before_script:
-  - docker-compose -f test/docker-compose.yml up -d zookeeper
-  - docker-compose -f test/docker-compose.yml scale kafka=2
+  - docker-compose -f test/docker-compose.yml up -d zookeeper kafka_1 kafka_2
 
 script:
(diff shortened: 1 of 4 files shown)


wurstmeister:
check for an alternative https://github.com/wurstmeister/kafka-docker/pull/538

sscaling:
I considered this, but then they seem fragile to future changes as well. This just makes everything explicit and reduces one step in the test. I can go either way on it if you have a strong preference.

wurstmeister:
i definitely think the compose version should be pinned to avoid issues with future breaking changes in compose. 
This change seems to make the compose file a little more verbose just to work around the change that has been reverted again.

sscaling:
Decided that we'll do this and #538

PR #608, opened 2020-07-03
Add a case sensitive option for adding entries to log4j.properties

jcaesar:
#188 / 5d324f14ccd7030355e618e81a31ef0d45dcba35 added reading the LOG4J_ environment variables, but converts all keys to lowercase.
Please correct me if I'm wrong on this: Matches in log4j.properties are case sensitive, this LOG4J_ only allowed configuring logging per package (since java package names are lower case), but not for single classes (which are in CamelCase).

I've decided to keep LOG4J_ for backwards compatibility. Also note that I think this PR is of minor importance, since the functionality of both log4j_ and LOG4J_ can be easily replaced by
```yaml
    entrypoint: |
      bash -c "
        echo >>/opt/kafka/config/log4j.properties log4j.logger.kafka.server.KafkaConfig=WARN
        exec start-kafka.sh
      "
```
and similar tricks.

### DIFF
--- a/start-kafka.sh
+++ b/start-kafka.sh
@@ -139,6 +139,11 @@ echo "" >> "$KAFKA_HOME/config/server.properties"
             log4j_name=$(echo "$env_var" | tr '[:upper:]' '[:lower:]' | tr _ .)
             updateConfig "$log4j_name" "${!env_var}" "$KAFKA_HOME/config/log4j.properties"
         fi
+
+        if [[ $env_var =~ ^log4j_ ]]; then
+            log4j_name=$(echo "$env_var" | tr _ .)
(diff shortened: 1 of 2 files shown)


jcaesar:
@gwilym as you are the original author, I'd be happy if I could get a comment from you.

gwilym:
Appreciate the mention @jcaesar however unfortunately I'm not sure I can be of much assistance as I'm not that familiar with how uses of log4j can vary. I only have surface awareness of log4j from an operational standpoint (e.g. reading documentation to run a containerised Kafka and similar services) and I wasn't aware that class-specific configurations like this were a feature.

So I'd have to defer to the Java / log4j experts 🙇

Felixoid:
Hey @wurstmeister, is there anything that would push this PR further?

jcaesar:
Time to let go. This has a good workaround anyway.

PR #640, opened 2021-01-07
Added the option to specify topic.retention.ms during topic creation

gontcharovd:
I have extended the `create-topics.sh` script to include the option to specify the retention (in milliseconds) during topic creation.
The retention is specified by extending the existing topic creation syntax: name:partitions:replicas:cleanup.policy:**retention.ms**.
If `topicConfig[4]` exists, the retention is set for that topic.


renatomefi:
Hey, I'm proposing something similar, check it out: https://github.com/wurstmeister/kafka-docker/pull/642

PR #655, opened 2021-04-20
Add support for Kafka version 2.8.0

YemoAde:
- Kafka Version 2.8.0 has first release of  KIP 500 (Removes dependence on zookeeper)
- However, current release is for development only because zookeeper is still required for Access Control List in Kafka
- See Kafka latest release log

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,18 +13,6 @@ env:
 # Build recommended versions based on: http://kafka.apache.org/downloads
 matrix:
   include:
-  - scala: "2.10"
-    env: KAFKA_VERSION=0.8.2.2
-  - scala: 2.11
-    env: KAFKA_VERSION=0.9.0.1
-  - scala: 2.11
-    env: KAFKA_VERSION=0.10.2.2
-  - scala: 2.11
-    env: KAFKA_VERSION=0.11.0.3
-  - scala: 2.11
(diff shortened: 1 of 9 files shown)


sycured:
I think that using an environment variable, we can change the startup between:

- with kraft
- with zookeeper

and having a default startup mode.

Just an idea to have a container supporting both modes.

.travis.yml
```diff
@@ -5,7 +5,7 @@ language: scala
 services:
   - docker
 
-# This version will be also tagged as 'latest'
+# This release is experimental only, previous version is remains 'latest'
```
wurstmeister, line 8: According to https://kafka.apache.org/downloads: `2.8.0 is the latest release. The current stable version is 2.8.0.`  so this comment can be reverted

alkmc:
@YemoAde any update on this subject?

.travis.yml
YemoAde: Fixed !
@wurstmeister kindly review this PR. Support for KRaft Mode without zookeeper should/will be a separate PR.

wurstmeister:
Thanks for your PR, it's been superseded by https://github.com/wurstmeister/kafka-docker/commit/85ca0564640b189f86c4f5842ab6d5fe0d80658e now

PR #656, opened 2021-04-22
fix broker-list.sh if two or more brokers

jackjliew:
According to the official doc below, "If pattern begins with ‘/’, all matches of pattern are replaced with string. Normally only the first match is replaced."

[3.5.3 Shell Parameter Expansion](https://www.gnu.org/software/bash/manual/html_node/Shell-Parameter-Expansion.html)


wurstmeister:
Thank you!

Issue #647, opened 2021-03-22
Can't connect to Kafka on Apple Silicon 


Crowelling:
Hello.

When i am trying to run kafka on apple with silicon m1 chip, i have no errors on logs, but can't connect to kafka container. Tried to use simple docker-compose.yml:

```version: '3'
services:
  zookeeper:
    image: wurstmeister/zookeeper
    ports:
      - "2181:2181"
  kafka:
    image: wurstmeister/kafka
    links:
      - zookeeper
    ports:
      - "9092:9092"
    environment:
      KAFKA_ADVERTISED_HOST_NAME: localhost
      KAFKA_ZOOKEEPER_CONNECT: zookeeper:2181
      KAFKA_CREATE_TOPICS: "test-topic:5:2"
    volumes:
      - /var/run/docker.sock:/var/run/docker.sock
 ```
 
No errors on logs, just telling that `waiting for kafka to be ready` and that's it. Can't connect to localhost:9092

Maybe somebody knows where is the problem?

shubham1g5:
I am facing the same issue.

jdamiani27:
wurstmeister/kafka currently does not provide arm builds and it is likely that the qemu emulation layer is failing (many users encountering segfaults in various amd64 images https://github.com/docker/for-mac/issues/5123)

The openjdk base image does provide arm builds so I've simply built my own image locally from this repo like so (additionally using the official zookeeper image with arm builds):

`version: '3'

services:
  zookeeper:
    image: zookeeper
    ports:
      - "2181:2181"

  kafka:
    build: ./kafka-docker
    ports:
      - "9092:9092"
    environment:
      KAFKA_ADVERTISED_HOST_NAME: localhost
      KAFKA_ZOOKEEPER_CONNECT: zookeeper:2181
    volumes:
      - /var/run/docker.sock:/var/run/docker.sock`

chetanyachopra:
> wurstmeister/kafka currently does not provide arm builds and it is likely that the qemu emulation layer is failing (many users encountering segfaults in various amd64 images [docker/for-mac#5123](https://github.com/docker/for-mac/issues/5123))
> 
> The openjdk base image does provide arm builds so I've simply built my own image locally from this repo like so (additionally using the official zookeeper image with arm builds):
> 
> ```
> version: '3'
> 
> services:
>   zookeeper:
>     image: zookeeper
>     ports:
>       - "2181:2181"
> 
>   kafka:
>     build: ./kafka-docker
[...]

PR #659, opened 2021-05-05
Update openjdk image

mangas:
The previous image was crashing on M1. After trying a few different
images, this one seems to solve the problem

Closes https://github.com/wurstmeister/kafka-docker/issues/647

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM openjdk:8u212-jre-alpine
+FROM azul/zulu-openjdk-alpine:8u292-8.54.0.21
 
 ARG kafka_version=2.7.0
 ARG scala_version=2.13
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,10 @@ Changelog
 
 Kafka features are not tied to a specific kafka-docker version (ideally all changes will be merged into all branches). Therefore, this changelog will track changes to the image by date.
 
+06-Jun-2021
+----------
+- Add support for darwin arm by to azul/zulu-openjdk-alpine base image
+
 05-Jun-2021
 -----------
 
--- a/download-kafka.sh
+++ b/download-kafka.sh
@@ -9,7 +9,7 @@ url=$(curl --stderr /dev/null "https://www.apache.org/dyn/closer.cgi?path=/kafka
 
 # Test to see if the suggested mirror has this version, currently pre 2.1.1 versions
 # do not appear to be actively mirrored. This may also be useful if closer.cgi is down.
-if [[ ! $(curl -s -f -I "${url}") ]]; then
+if [[ ! $(curl -f -s -r 0-1 "${url}") ]]; then
     echo "Mirror does not have desired version, downloading direct from Apache"
     url="https://archive.apache.org/dist/kafka/${KAFKA_VERSION}/${FILENAME}"
 fi


Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM openjdk:8u212-jre-alpine
+FROM azul/zulu-openjdk-alpine:16.0.1-16.30.15-jre
```
jaceklaskowski, line 1: That does not seem as a mere replacement (to fix an issue with M1) but a version upgrade (that impacts all the other non-M1 systems). Is this intentional?

mangas: It's intentional since version 8 was actually hanging and therefore an upgrade was needed. From my testing jdk11 had the same problem

mangas: I have updated the image now to a java8 based image that works with the m1

mangas:
@huang-x-h could you please review this since you seem to have been the original author

mjeri:
Can confirm, this works also in my setup on the M1 👍 Thanks!

huang-x-h:
> @huang-x-h could you please review this since you seem to have been the original author

Sorry, the author is @wurstmeister , i just upgrade the kafka version to 2.7.0

wurstmeister:
Hi, 

this change breaks the build. 

- upgrades alpine from v3.9 - v3.13
- upgrades curl 7.64.0-r5 - (7.76.1-r0) 

The problem seems to be a change in behaviour of curl 

with the new version of curl 

```

**Comment by mangas** [comment]
I'm not sure what the previous comment is referring to, is it a build for a previous version of kafka? 

```
 docker build --no-cache  .
[+] Building 29.0s (9/9) FINISHED
 => [internal] load build definition from Dockerfile                                                                                                                                                            0.0s
 => => transferring dockerfile: 37B                                                                                                                                                                             0.0s
 => [internal] load .dockerignore                                                                                                                                                                               0.0s
 => => transferring context: 2B                                                                                                                                                                                 0.0s
 => [internal] load metadata for docker.io/azul/zulu-openjdk-alpine:8u292-8.54.0.21                                                                                                                             0.6s
 => CACHED [1/4] FROM docker.io/azul/zulu-openjdk-alpine:8u292-8.54.0.21@sha256:86424b902f10c1e8f1484e6e2b79f2c9e0eaf0be9bbdf8f4933e49b5f2521980                                                                0.0s
 => [internal] load build context                                                                                                                                                                               0.0s
 => => transferring context: 243B                                                                                                                                                                               0.0s
 => [2/4] COPY download-kafka.sh start-kafka.sh broker-list.sh create-topics.sh versions.sh /tmp/                                                                                                               0.1s
 => [3/4] RUN apk add --no-cache bash curl jq docker  && chmod a+x /tmp/*.sh  && mv /tmp/start-kafka.sh /tmp/broker-list.sh /tmp/create-topics.sh /tmp/versions.sh /usr/bin  && sync && /tmp/download-kafka.s  27.5s
 => [4/4] COPY overrides /opt/overrides                                                                                                                                                                         0.0s
 => exporting to image                                                                                                                                                                                          0.7s
 => => exporting layers                                                                                                                                                                                         0.7s
 => => writing image sha256:98e803fd50b188bdf31d360790d89deefedb8a1ce8b8fbc3189ed4ebc3880799
```

mangas:
Also there seems to be a rate limit problem with the build at the moment "Error response from daemon: toomanyrequests: You have reached your pull rate limit. You may increase the limit by authenticating and upgrading: https://www.docker.com/increase-rate-limit"

wurstmeister:
Thanks for updating the PR @mangas, could you please also update the changelog?

regarding the rate limit, I need to check what we can do to address that.

mangas:
@wurstmeister updated the changelog as well. Did you have a change to look into the rate limit issue? Do you think we could merge this in?

wurstmeister:
@mangas thanks for the update. I've made a couple changes to reduce the chance of the rate limit issue happening, at the very least when building the master branch. could you please update your branch with the latest changes, then we can merge your change.

mangas:
@wurstmeister seems like we're still being rate limited but I've rebased as requested

mangas:
Perhaps making the build steps sequencial could help with the rate limiting problem?

PR #674, opened 2021-07-08
Fix sed: unmatched '@'

LatentLag:
When a container is restarted `docker port` reports multiple ports which yields the error `sed: unmatched '@'` because there's a newline in the sed result. This change picks the first port, which removes the first line.


KielD-01:
@wurstmeister any chance to have this merged?

wurstmeister:
Thank you!

PR #684, opened 2021-09-03
build issue

christianrua:
when you use the command "docker-compose -f docker-compose-single-broker.yml up", the docker raises a error, but with this fix, it works.

### DIFF
--- a/docker-compose-single-broker.yml
+++ b/docker-compose-single-broker.yml
@@ -5,7 +5,7 @@ services:
     ports:
       - "2181:2181"
   kafka:
-    build: .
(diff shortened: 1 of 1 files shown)


OneCricketeer:
What is the error? This pulls from DockerHub, not builds the code in the repo, so you'd potentially be pulling an old image and not using any code from what you've cloned other than the compose file itself

christianrua:
that ok, i am not very familiar with some things, i will try to run the repo again in local mode, and i will try to figure out how works or what is the correct use of this repo. thanks for the answer.

Issue #699, opened 2022-01-06
Build not stable in Apple Silicon


bugflux:
I understand this may not be a very informative way to put it, I'm happy to provide more info if necessary.

I've noticed that neither `wurstmeister/kafka:latest` or `wurstmeister/kafka:2.13-2.8.1` are very stable under Apple Silicon chips (M1 Pro 32GB of RAM).

Compared with [this magical build](https://hub.docker.com/layers/schrbr/kafka/2.13-2.6.0/images/sha256-a1cf975f1aeac61b72bcaf73c7814fb3730f69e9917e8c6fae400c6d6705c004?context=explore):
- The CPU spikes quite a lot
  - It doesn't seem to be an issue with memory, it's the same with Xmx256m or 512m
- It takes longer to start
- It seems to hang occasionally

Any ideas what might be the cause, or what I can do or provide to find the causes?

bugflux:
It seems to run fine if building `FROM bellsoft/liberica-openjdk-alpine:11`.

rubensayshi:
I think we're seeing similar issues, or issues when the macbook comes out of a hibernate ...

OACDesigns:
I'm also encountering strange issues with Kafka running in docker on Apple Silicon (also M1 Pro 32GB RAM).
I'm using `wurstmeister/kafka:2.13-2.8.1` and getting strange crashes here and there. I tell  the cause of them, and don't know how to reproduce. But it's happened several times now.
 
The latest one resulted in a dump to screen which I've saved to this gist: https://gist.github.com/OACDesigns/580e0f816a96b4bd292cdca5ebddeb4e

Maybe someone is better at understanding the logs than I am.

jimbogithub:
I had similar issues trying the wurstmeister/kafka:2.13-2.7.1.  Looks like although it claims to be an arm64 image:

```
$ docker image inspect wurstmeister/kafka:2.13-2.7.1

[
    {
        "Id": "sha256:e0c8865c03ee43e3247476d85459c6d3009da6395b951752af7d750ca5a976ac",
        "RepoTags": [
            "wurstmeister/kafka:2.13-2.7.1"
        ],
...
        "Architecture": "arm64",
        "Os": "linux",
...
```

it's actually amd64 inside:

```
bash-5.1# uname -a
Linux 29e98d5a58d0 5.10.76-linuxkit #1 SMP PREEMPT Mon Nov 8 11:22:26 UTC 2021 x86_64 Linux
[...]

Issue #544, opened 2019-11-11
Support for Kafka on Java 11


amuraru:
Kafka supports now java-11 https://issues.apache.org/jira/browse/KAFKA-7264.
As java-11 brings better ergonomics when running in docker I think it would be useful to have java-11 supported in this image

amuraru:
I did a quick test in my environment and switched to `adoptopenjdk/openjdk11:x86_64-alpine-jre-11.0.5_10` base image

https://github.com/wurstmeister/kafka-docker/compare/master...amuraru:java11#diff-3254677a7917c6c01f55212f86c57fbf

This seems to work.
Would it make sense to add support for java11 in https://github.com/wurstmeister/kafka-docker/pull/536 for example?

sscaling:
Java 11 isn't supported on older versions https://kafka.apache.org/downloads. It has only been fully supported since 2.1.0. So to support this the change isn't as trivial as you propose. It would at a minimum require the base image being configurable per version.

PR #703, opened 2022-02-22
Switch from Alpine to Debian to get newer multi-arch JRE

jimbogithub:
See https://github.com/wurstmeister/kafka-docker/issues/699 for prior discussion.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,8 +14,6 @@ env:
 matrix:
   include:
   - scala: 2.12
-    env: KAFKA_VERSION=2.0.1
-  - scala: 2.12
     env: KAFKA_VERSION=2.1.1
   - scala: 2.12
     env: KAFKA_VERSION=2.2.2
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,12 @@ Changelog
 
 Kafka features are not tied to a specific kafka-docker version (ideally all changes will be merged into all branches). Therefore, this changelog will track changes to the image by date.
 
+09-Apr-2022
+----------
+
+-	Switched to openjdk:11-jre-slim for multi-arch support. 
+-	Drop support for Kafka `2.0.1` due to JDK upgrade.
+
 04-Oct-2021
 ----------
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,8 +1,7 @@
-FROM azul/zulu-openjdk-alpine:8u292-8.54.0.21
+FROM openjdk:11-jre-slim
 
 ARG kafka_version=2.8.1
 ARG scala_version=2.13
-ARG glibc_version=2.31-r0
 ARG vcs_ref=unspecified
 ARG build_date=unspecified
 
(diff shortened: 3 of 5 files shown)


jimbogithub:
This is about as far as I can get.  The two `### BEGIN .... ### END` blocks show the extra packages that had to be installed to make progress in the CI build.  As far as my tests have found, none of that is necessary for the image to work correctly otherwise.  Perhaps someone could refactor to use a derived image for the tests?

I also added `libsnappy1v5` to the main list of packages in case that was related to the broken test reported as `compact-1645575814 topic not configured correctly: 'compact-1645575814:compact'` but it doesn't make any difference so I'm not sure if it's really needed.  It seems to me that the test may be wrong...

I tried something similar to the failing test on a local container:

```
kafka-topics.sh --bootstrap-server $KAFKA_BOOTSTRAP_SERVERS --create --topic CompactTest --partitions 1 --config cleanup.policy=compact --config=compression.type=snappy --replication-factor 3

kafka-configs.sh --bootstrap-server $KAFKA_BOOTSTRAP_SERVERS --entity-type topics --entity-name CompactTest --describe 

Topic: CompactTest	TopicId: KxUmDJ3IQg2gku8Nos7jnQ	PartitionCount: 1	ReplicationFactor: 3	Configs: compression.type=snappy,cleanup.policy=compact,segment.bytes=1073741824,max.message.bytes=2000000
	Topic: CompactTest	Partition: 0	Leader: 1001	Replicas: 1001,1003,1002	Isr: 1001,1003,1002
```

The test uses this command to confirm settings:
```
kafka-configs.sh --bootstrap-server $KAFKA_BOOTSTRAP_SERVERS --entity-type topics --entity-name CompactTest --describe | grep 'Configs for topic' | awk -F'cleanup.policy=' '{print $2}'
```

If I do the first bit I see good values:
```
kafka-configs.sh --bootstrap-server $KAFKA_BOOTSTRAP_SERVERS --entity-type topics --entity-name CompactTest --describe
Dynamic configs for topic CompactTest are:
  compression.type=snappy sensitive=false synonyms={DYNAMIC_TOPIC_CONFIG:compression.type=snappy, DEFAULT_CONFIG:compression.type=producer}
  cleanup.policy=compact sensitive=false synonyms={DYNAMIC_TOPIC_CONFIG:cleanup.policy=compact, DEFAULT_CONFIG:log.cleanup.policy=delete}
```
 But if you include the `grep`, e.g:
```
kafka-configs.sh --bootstrap-server $KAFKA_BOOTSTRAP_SERVERS --entity-type topics --entity-name CompactTest --describe | grep 'Configs for topic'
```
then the result is blank and the test fails.

jimbogithub:
OK, building after a few more whacks with the hammer.

- Are the changes to the test OK?
- Having docker there just for testing is adding 150MB to the image size, but that's how it was before.

wurstmeister:
Thanks for putting this together. 

- this change also switches from java 8 to java 11, as mentioned in https://github.com/wurstmeister/kafka-docker/issues/544#issuecomment-568071737, java 11 has only been officially supported since 2.1.0 (https://issues.apache.org/jira/browse/KAFKA-7264, https://archive.apache.org/dist/kafka/2.1.0/RELEASE_NOTES.html) so we should drop support for 2.0.1 (https://github.com/wurstmeister/kafka-docker/blob/master/.travis.yml#L17)

tednaleid:
Confirming that #703 is much more stable for me on my M1 macs than the current `latest` release.  Are there any blockers on merging #703 to make that support part of latest?

Is the PR from sfat above enough to get this merged, or are there other things that would need to be done?

jimbogithub:
This will also work with `openjdk:17-jdk-slim` which appears to be supported (https://cwiki.apache.org/confluence/pages/viewpage.action?pageId=181308223 - _we support the two most recent LTS versions and the most recent non LTS version_) but it further bloats the image due to being a full JDK rather than JRE.  If we care about the image size I'd be happy to make the JRE if someone else wanted to help figure out how to remove the embedded Docker required for CI tests.  Do we know why the embedded Docker is there?  Is it just to get the list of ports?

jimbogithub:
@wurstmeister Can this be merged now?

wurstmeister:
Thank you!

TBragi:
Just a heads up, this change from Apline to Debian gave me an error:
I am using a setup where the hostname of my kafka cluster nodes is something in the likes of:
```
server01-domain1.com
server02-domain1.com
server03-domain1.com
```

I had previously used environment variable `BROKER_ID_COMMAND: 'hostname | grep -oE "(\d+)" | head -1'`
which resulted in broker ID's: `01`, `02`, `03`.

After this merge the grep would catch the `d` in `domain` instead. This has been fixed in two ways:
* `BROKER_ID_COMMAND: 'hostname | grep -oE "[0-9]+" | head -1'`  this uses "correct" Extended regular expression
* `BROKER_ID_COMMAND: 'hostname | grep -oP "(\d+)" | head -1'` this uses Perl Regular expression

It seems like the previous grep had the Perl RegEx as part of the Extended RegEx, but that is no longer the case?

PR #706, opened 2022-04-27
Updating base image to tag 8u332-8.62.0.19 for security updates

Raimmaster:
The current base image uses Alpine v3.4.6. This version of Alpine can't update all of its packages if they are suffering from vulnerabilities. Updating to `azul/zulu-openjdk-alpine:8u332-8.62.0.19` will allow it to use Alpine v3.15 without affecting the rest of functionality, and allowing Dockerfiles based on this kafka-docker image to upgrade their packages to patched versions without affecting this one.


wurstmeister:
this is not required anymore after https://github.com/wurstmeister/kafka-docker/pull/703

PR #710, opened 2022-05-13
Adjust docker-compose single example

michalhosala:
Using KAFKA_ADVERTISED_HOST_NAME is now deprecated: https://kafka.apache.org/0110/documentation.html#brokerconfigs

### DIFF
--- a/docker-compose-single-broker.yml
+++ b/docker-compose-single-broker.yml
@@ -9,7 +9,8 @@ services:
     ports:
       - "9092:9092"
     environment:
-      KAFKA_ADVERTISED_HOST_NAME: 192.168.99.100
+      KAFKA_ADVERTISED_LISTENERS: PLAINTEXT://localhost:9092
+      KAFKA_LISTENERS: PLAINTEXT://:9092
       KAFKA_CREATE_TOPICS: "test:1:1"
       KAFKA_ZOOKEEPER_CONNECT: zookeeper:2181
     volumes:


docker-compose-single-broker.yml
```diff
@@ -9,7 +9,8 @@ services:
     ports:
       - "9092:9092"
     environment:
-      KAFKA_ADVERTISED_HOST_NAME: 192.168.99.100
+      KAFKA_ADVERTISED_LISTENERS: PLAINTEXT://localhost:9092
+      KAFKA_LISTENERS: PLAINTEXT://:9092
```
OneCricketeer, line 13: By default, it'll only bind internally to the compose network 

```suggestion
      KAFKA_LISTENERS: PLAINTEXT://0.0.0.0:9092
```

michalhosala: I am no Kafka expert, but this change should not be necessary as exposure to the outside world via `localhost:9092` is assured by `KAFKA_ADVERTISED_LISTENERS` on the above line, see [docs](https://kafka.apache.org/documentation/#brokerconfigs_advertised.listeners).

OneCricketeer: This variable sets the bind address of the Java server. The advertised listeners are what's stored in Zookeeper and what are returned to clients for future connections. They are very different settings

OneCricketeer: Also worth mentioning that setting the advertised listener to localhost will not allow you to add other Kafka client services to the compose network, as they will then try to connect to themselves rather than the broker container. Perhaps you should add two listeners

OneCricketeer: > it is available within compose as well thanks to KAFKA_LISTENERS at kafka:9092.

The bootstrap is. The actual client protocol is not. 

Try this

```
$ docker-compose up -d
$ docker network ls  # get the network
$ docker run --rm -ti --entrypoint=bash --network=<name> wurstmeister/kafka
root@41cab124273d:/# nc -vz kafka 9092
Connection to kafka (172.19.0.2) 9092 port [tcp/*] succeeded!
root@41cab124273d:/# hostname -f
41cab124273d
root@41cab124273d:/# /opt/kafka/bin/kafka-topics.sh --list --bootstrap-server kafka:9092
[2022-05-13 17:04:58,904] WARN [AdminClient clientId=adminclient-1] Connection to node 1001 (localhost/127.0.0.1:9092) could not be established. Broker may not be available. (org.apache.kafka.clients.NetworkClient)
```

michalhosala: > The bootstrap is. The actual client protocol is not.

but the proposed change (`KAFKA_LISTENERS: PLAINTEXT://0.0.0.0:9092`) is not solving it, is it? It is likely my limited Kafka knowledge, but what s the addition of 0.0.0.0 good for here?

The steps you provided in you last post demonstrate the issue with `localhost`, not why the `0.0.0.0` should be added, right?

OneCricketeer: You're correct that I'm proposing a different solution. You've set the value [as the default](https://docs.confluent.io/platform/current/installation/configuration/broker-configs.html#brokerconfigs_listeners), and so it'll bind to the container's canonical hostname. Since Compose allows containers over the bridge network to address hosts using service names, then it'll continue to work, but also, as linked setting to `0.0.0.0` "binds to all interfaces", which is generally more appropriate for allowing external connections.

andrewchoi5 (approved):

